PR #3, opened 2019-05-07
codebuild support

kapilt:
for https://github.com/ansible/ansible/pull/47187

it looks good to me, but i'm gonna work on a simple cli runner for easier manual testing on terminator classes.


kapilt:
so there's an open question on the iam for this so moving back to wip, the originating pr, was actually hit an issue unrelated to codebuild support in terminator because its actual pain point was bad prefix on its role.

https://github.com/ansible/ansible/pull/47187

PR #4, opened 2019-05-07
minor refactor on session passing and resource processing

kapilt:
- first step is was switching raw credentials being weaved through calls to boto3.session


kapilt:
the clirunner removed, i archived it over here in case anyone else i interested.. its main value is being able to test/check resources individually.

https://gist.github.com/kapilt/85497259de49ea54fc232b451cc6a398

PR #5, opened 2019-05-07
aws - codepipeline support

kapilt:
for https://github.com/ansible/ansible/pull/47189

PR #41, opened 2019-06-10
Added CodePipeline support

stefanhorning:
Based on PR #5

### DIFF
--- a/aws/terminator/application_services.py
+++ b/aws/terminator/application_services.py
@@ -82,6 +82,31 @@ class CodeCommitRepository(DbTerminator):
         self.client.delete_repository(repositoryName=self.name)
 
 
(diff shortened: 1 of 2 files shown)


aws/terminator/application_services.py
```diff
@@ -82,6 +82,31 @@ def terminate(self):
...
+
+    @property
+    def id(self):
+        return self.instance['name']
+
+    @property
+    def name(self):
+        return self.instance['name']
+
+    @property
```
mattclay:
```suggestion
```

mattclay:
It looks like one minor change is needed (unnecessary property), then this should be good to merge.

stefanhorning:
Fixed

PR #44, opened 2019-07-11
Update s3 bucket policies

aljazkosir:
- Add s3:GetBucketNotification permission
- Add s3:PutBucketNotification permission


mattclay:
For https://github.com/ansible/ansible/pull/58059

PR #47, opened 2019-08-12
Add EKS support

jillr:
Pre-requisite for enabling EKS tests.

### DIFF
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -31,8 +31,6 @@ Statement:
           - arn:aws:iam::aws:policy/IAMReadOnlyAccess
           - arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM
           - arn:aws:iam::aws:policy/service-role/AmazonEC2ContainerServiceRole
-          - arn:aws:iam::aws:policy/AmazonEKSClusterPolicy
-          - arn:aws:iam::aws:policy/AmazonEKSServicePolicy
   - Sid: AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees
     Effect: Allow
     Action:
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -31,6 +31,7 @@ Statement:
           - arn:aws:iam::aws:policy/IAMReadOnlyAccess
           - arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM
           - arn:aws:iam::aws:policy/service-role/AmazonEC2ContainerServiceRole
+          - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
   - Sid: AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees
     Effect: Allow
     Action:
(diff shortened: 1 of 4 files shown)


aws/policy/security-services.yaml
```diff
@@ -28,6 +28,8 @@ Statement:
           - arn:aws:iam::aws:policy/IAMReadOnlyAccess
           - arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM
           - arn:aws:iam::aws:policy/service-role/AmazonEC2ContainerServiceRole
+          - arn:aws:iam::aws:policy/AmazonEKSClusterPolicy
```
s-hertel: In general this looks great. It would be good to get @mattclay's opinion on allowing this policy to be attached to roles. Here's an example where the same underlying issue this policy exposes has been deemed not to be a viable option- https://github.com/ansible/ansible-core-ci/pull/197#discussion_r261330886

An alternative would be to create a static role for the CI account, and only try to create the role/attach the policy if it does not already exist. Then users could still have the benefit of not having to create the role for their account ahead of time, but the CI account would not need this permission.

mattclay: I agree that the static role is our only option currently, until we're able to run the terminator in all regions.

jillr: Works for me. AFAICT that's not something I can do through this repo though, unless I'm missing some breadcrumb the CreateSecret example should lead me too.  How do I create a static role?

jillr: @s-hertel @mattclay can either of you point me to how I create the static role?  I'm pretty sure I don't have privs to do so in the console as I can't see permissions for any existing roles and none of the roles I can see look like they'd cover the noted example.

s-hertel: Sorry, @jillr. I already created the role when we talked about it in Slack last week. The role name is test-eks-role (to keep the naming pattern consistent), in case you forgot.

s-hertel:
Oh - actually. This looks great, but can you add terminator classes for the things the test-eks-role service role could create?

jillr:
@s-hertel  Like ELBs and target groups right (or anything else I haven't thought of yet)?  Yeah I have another branch already WIP on that that I can pull into this one, the target groups are a bit fiddly so it'll be a minute.

s-hertel:
@jillr Yeah, just those two I think. I just went through the policies to check.

jillr:
This is currently deployed to stage=dev and I've been testing with https://github.com/ansible/ansible/pull/61256.
@s-hertel can you please re-review with the additional resources?

aws/policy/security-services.yaml
```diff
@@ -22,12 +22,14 @@ Statement:
       - iam:DetachRolePolicy
     Resource:
       - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-sts-*
+      - arn:aws:iam::{{ aws_account_id }}:role/ansible_lambda_role
```
s-hertel: What's this for and the change below?

jillr: I'm going to have to go with "something I was trying for some reason and forgot to remove". `aws_eks` passes on dev without these so away they go.

PR #48, opened 2019-08-14
Add state function service policies and terminator for PR59116

jillr:


aws/terminator/application_services.py
```diff
@@ -302,3 +302,26 @@ def name(self):
...
         return self.client.delete_table(TableName=self.instance)
+
+
+class StepFunctions(Terminator):
+    @staticmethod
+    def create(credentials):
+
+        def get_state_machines(client):
+            state_machines = client.get_paginator(
+                'list_state_machines').paginate().build_full_result().get('stateMachines', ())
```
s-hertel: Not a blocker, but I generally try to just have one return type for a given function.
```suggestion
                'list_state_machines').paginate().build_full_result().get('stateMachines', [])
```

PR #50, opened 2019-08-26
Add a how-to for adding terminator classes and testing them

s-hertel:


aws/README.md
```diff
@@ -1,3 +1,93 @@
+# Contributor process
+
+If you have a pull request that contains tests that need to be enabled in CI you can get the ball rolling by
```
matt-telstra: > If you have a pull request for [Ansible](https://github.com/ansible/ansible/) which contains tests that need to be enabled in CI you can get the ball rolling by
submitting a pull request to [this repository](https://github.com/matt-telstra/aws-terminator/). Start by updating the appropriate test policy in the policy directory (policy
groups are defined below).

aws/README.md
```diff
@@ -1,3 +1,93 @@
...
+Using cleanup.py with your own account:
+  - Ensure you're using Python 3.7+
+  - Modify config.yml to use your own accounts. These can be the same account if you're just using cleanup.py.
+    If you are using two separate accounts, lambda_account_id should be the account of the profile that's going
+    to assume the IAM role in the test_account_id. The test_account_id is where the terminator class(es) will run.
+  - Make sure you have a role called ansible-core-ci-test-dev that your AWS profile can assume. This role should have
+    the permissions to use your cleanup.py class.
+  - Set the environment variable AWS_PROFILE with the profile you want to use.
+  - Run the cleanup.py script for the new target you added to find any stray resources in us-east-1: |
+        $ python cleanup.py --stage dev --target Ec2Instance -v -c
```
matt-telstra: The Github Markdown renderer is not showing this as code. Maybe you need an empty line before the code starts?

aws/README.md
```diff
@@ -1,3 +1,93 @@
...
+    @staticmethod
+    def create(credentials):
+
+        def get_instances(client):
+            return [i for r in client.describe_instances()['Reservations'] for i in r['Instances']]
+
+        return Terminator._create(credentials, Ec2Instance, 'ec2', get_instances)
+```
+
+The `id`, `name`, and `created_time` properties should return an appropriate value from self.instance which is
```
matt-telstra: Is it acceptable for `name` to not be unique? (e.g. two ACM certificates with the same domain, two Route53 hosted zones with the same domain, etc)

matt-telstra: Looking at existing code, it seems the answer is that they don't have to be unique. So maybe state that explicitly here?

PR #53, opened 2019-09-04
Rename role assumption policy 

jillr:
Rename role assumption policy as it's being used for non-sts tests now
Add role/ansible-test-* resources to also reflect this

@tremble makes a good point in https://github.com/ansible/ansible/pull/61256 that we've started overloading use of arn:aws:iam::{{ aws_account_id }}:role/ansible-test-sts-* for tests that are not related to STS.

### DIFF
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -112,3 +112,4 @@ Statement:
     Resource:
       - arn:aws:iam::{{ aws_account_id }}:role/ansible_lambda_role
       - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-sts-*
+      - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-*
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -1,6 +1,6 @@
 Version: '2012-10-17'
 Statement:
-  - Sid: AllowStsAssumeRoleTests
+  - Sid: AllowAssumeRoleTests
     Effect: Allow
     Action:
       - iam:CreateRole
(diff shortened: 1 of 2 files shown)


jillr:
cc @s-hertel

s-hertel:
This should be updated for iam:PassRole too!

jillr:
> This should be updated for iam:PassRole too!

I should know better than to push right before lunch - thanks!  :)

aws/policy/security-services.yaml
```diff
@@ -15,13 +15,15 @@ Statement:
     Resource:
       - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-sts-*
       - arn:aws:iam::{{ aws_account_id }}:instance-profile/ansible-test-sts-*
-  - Sid: AllowStsAssumeRoleTestsAttachAndDetachRole
+      - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-*
```
s-hertel: I'm not sure if it was intentional or not, but do you want to remove the ansible-test-sts-* lines?
`arn:aws:iam::{{ aws_account_id }}:role/ansible-test-sts-*` should be covered by this line. You could also add `- arn:aws:iam::{{ aws_account_id }}:instance-profile/ansible-test-*`. Besides that, looks good to me.

jillr: I was erring on the side of leaving things very explicit, in case we did decide to break things down differently at some point we wouldn't overlook the different prefixes in use. I don't feel especially strongly either way so whatever seems sanest to others who need to work with the policies works for me.

s-hertel: I think they should be removed, but for the time being until we sort out the tests, looks good to me.

aws/policy/security-services.yaml
```diff
@@ -15,13 +15,15 @@ Statement:
     Resource:
       - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-sts-*
       - arn:aws:iam::{{ aws_account_id }}:instance-profile/ansible-test-sts-*
```
s-hertel: (Oh, oops, this is still needed for ec2_instance tests though if they switch from using ansible-test-sts. Might as well add that now.)
```suggestion
       - arn:aws:iam::{{ aws_account_id }}:instance-profile/ansible-test-sts-*
       - arn:aws:iam::{{ aws_account_id }}:instance-profile/ansible-test-*
```
Let me know when you want to deploy to prod, looks good to me besides that.

jillr: @s-hertel  Good to go - want me to do that or do you already have things in flight?

s-hertel: @jillr Go for it!

PR #61, opened 2019-10-01
Allow m1.large instance type

Shaps:
This is needed to allow the tests in [ansible/ansible#61222](https://github.com/ansible/ansible/pull/61222) to run (I have the updated tests and the fixed sanity checks changes locally, waiting on this to be merged before pushing).

I know the instance is quite a big one, but unfortunately, as you can see from the list, the auto ebs_optimization only works for these instance types, and m1.large seems to be the cheapest one.

/cc @jillr

### DIFF
--- a/aws/policy/compute.yaml
+++ b/aws/policy/compute.yaml
@@ -68,6 +68,7 @@ Statement:
           - t2.micro
           - t3.nano
           - t3.micro
+          - m1.large
   - Sid: AllowEc2RunInstances
     Effect: Allow
     Action:
--- a/aws/policy/compute.yaml
+++ b/aws/policy/compute.yaml
@@ -68,7 +68,7 @@ Statement:
           - t2.micro
           - t3.nano
           - t3.micro
-          - m1.large
+          - m1.large  # lowest cost instance type with EBS optimization supported
   - Sid: AllowEc2RunInstances
     Effect: Allow
     Action:


Shaps:
CI failure doesn't seem related to the change

aws/policy/compute.yaml
```diff
@@ -68,6 +68,7 @@ Statement:
           - t2.micro
           - t3.nano
           - t3.micro
+          - m1.large
```
mattclay: Is there a reason we can't use `t3a.nano` instead? It is listed as EBS-optimized by default:

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html

mattclay: Also, I see `t3.nano` is also listed there, in a second table under the same heading, although it does cost slightly more than the `t3a.nano` instances.

Shaps: It's because the instances listed in that table have EBS enabled by default and disabling it, won't have any effect.

> There is no need to enable EBS optimization and no effect if you disable EBS optimization

The instances we need to test are the ones listed in the **EBS Optimization Supported** table

mattclay: Thanks for the clarification. It looks like all the EBS Optimization Supported instances are previous generation instances. Do you know where the pricing is for those?

Shaps: Yeah, you can find them here: https://aws.amazon.com/ec2/previous-generation/

mattclay: Ah, I had looked there, but didn't see `m1.large` because I was looking at the wrong region.

It does appear to be the lowest cost, at $0.175/hour in us-east-1.

mattclay:
```suggestion
          - m1.large  # lowest cost instance type with EBS optimization supported
```

PR #63, opened 2019-10-03
Add permissions and termination for cloudwatch alarms

ezmac:
Permissions needed for https://github.com/ansible/ansible/pull/62669

cloudwatch alarms don't directly track their creation date.  Using boto3's describe_alarm_history will return history that might contain the created history item.  Alarm history is limited, so the create item might not be available.

### DIFF
--- a/aws/policy/application-services.yaml
+++ b/aws/policy/application-services.yaml
@@ -137,6 +137,13 @@ Statement:
       - logs:CreateLogGroup
       - logs:DeleteLogGroup
     Resource: arn:aws:logs:us-east-1:{{ aws_account_id }}:log-group:ansible-test*
+  - Sid: ModifyCloudwatchAlarms
+    Effect: Allow
+    Action:
+      - cloudwatch:PutMetricAlarm
+      - cloudwatch:DeleteAlarms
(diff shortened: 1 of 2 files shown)


aws/policy/application-services.yaml
```diff
@@ -130,6 +130,13 @@ Statement:
       - logs:CreateLogGroup
       - logs:DeleteLogGroup
     Resource: arn:aws:logs:us-east-1:{{ aws_account_id }}:log-group:ansible-test*
+  - Sid: ModifyCloudwatchAlarms
+    Effect: Allow
+    Action:
+      - cloudwatch:PutMetricAlarm
+      - cloudwatch:DeleteAlarms
+      - cloudwatch:DescribeAlarms
```
jillr: `cloudwatch:DescribeAlarms` has to be able to operate on `Resource: arn:aws:cloudwatch:us-east-1:{{ aws_account_id }}:alarm:*` to retrieve names. Could you please move this action to a `DescribeCloudwatchAlarms` sid that operates on that resource?

ezmac: Good catch.  Thanks for the review

PR #66, opened 2019-10-17
needs_ci_update backlog policies

jillr:
Add missing conditional to support ansible/ansible#61256
Add DynamoDB UpdateTable for ansible/ansible#45402
Allow disassociating IAM profiles for ansible/ansible#59152

Roll up a few very small changes to address backlogged PRs

### DIFF
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -12,7 +12,6 @@ Statement:
       - iam:DeleteInstanceProfile
       - iam:AddRoleToInstanceProfile
       - iam:RemoveRoleFromInstanceProfile
-      - iam:DisassociateIamInstanceProfile
     Resource:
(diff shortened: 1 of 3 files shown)


aws/policy/security-services.yaml
```diff
@@ -12,6 +12,7 @@ Statement:
       - iam:DeleteInstanceProfile
       - iam:AddRoleToInstanceProfile
       - iam:RemoveRoleFromInstanceProfile
+      - iam:DisassociateIamInstanceProfile
```
s-hertel: I don't see this one in the [docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_identityandaccessmanagement.html), only for `ec2:DisassociateIamInstanceProfile`.

PR #68, opened 2019-10-23
Update policy for lightsail testing

prasadkatti:
A PR for lightsail integration tests needs an update to the CI account.
Link to PR - https://github.com/ansible/ansible/pull/63770/

This is my first time submitting a patch here. So any help is greatly appreciated.

The `lightsail:CreateInstances` is the only action that incurs fees. So I have added that as a separate statement in the policy. Also, instances can only be created in us-east-1. I think that is how the other statements are written, so I tried to follow suit.

I wanted to add a condition so that only certain bundles could be used, but I couldn't find a way to do that.

### DIFF
--- a/aws/policy/compute.yaml
+++ b/aws/policy/compute.yaml
@@ -33,6 +33,12 @@ Statement:
       - ec2:StopInstances
       - ec2:TerminateInstances
       - ec2:UnassignPrivateIpAddresses
+      - lightsail:DeleteInstance
+      - lightsail:GetInstance
+      - lightsail:GetInstances
+      - lightsail:RebootInstance
+      - lightsail:StartInstance
+      - lightsail:StopInstance
     Resource:
       - "*"
     Condition:
@@ -181,3 +187,9 @@ Statement:
       - eks:CreateCluster
     Resource:
       - arn:aws:eks:us-east-1:{{ aws_account_id }}:cluster/*
+  - Sid: AllowLightsailCreateInstances
+    Effect: Allow
+    Action:
+      - lightsail:CreateInstances
+    Resource:
+      - arn:aws:lightsail:us-east-1:{{ aws_account_id }}:*
--- a/aws/policy/compute.yaml
+++ b/aws/policy/compute.yaml
@@ -33,12 +33,6 @@ Statement:
       - ec2:StopInstances
       - ec2:TerminateInstances
       - ec2:UnassignPrivateIpAddresses
-      - lightsail:DeleteInstance
-      - lightsail:GetInstance
-      - lightsail:GetInstances
-      - lightsail:RebootInstance
-      - lightsail:StartInstance
-      - lightsail:StopInstance
     Resource:
       - "*"
     Condition:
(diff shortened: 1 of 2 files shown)


jillr:
Thanks @prasadkatti, as Lightsail is not currently supported at all in the CI environment we will also need a new terminator class added.  The terminator is a lambda function we deploy that automatically destroys AWS resources after a certain age, in case a test or AWS service fails in a such a way that leaves resources behind.  There's a how-to in https://github.com/mattclay/aws-terminator/blob/master/aws/README.md if you would be able to take that on.

prasadkatti:
@jillr - I have added a terminator class for lightsail. I was able to test it too. I set the age limit to 2 mins for testing purposes and this is the output I get -

```
[INFO] 2019-10-27T18:17:28.222Z 1d81bedb-b2f0-472f-8dab-b439e0628723 skipped Lightsail: name=Amazon_Linux-2, age=0:00:12.675000, stale=False
[INFO] 2019-10-27T18:17:29.48Z 1d81bedb-b2f0-472f-8dab-b439e0628723 terminated Lightsail: name=Amazon_Linux-1, age=0:02:54.109000, stale=True
```

jillr:
Thanks @prasadkatti, the code looks good I'll be doing test deploys on new terminator changes this Thursday and will queue this PR for then.

aws/policy/compute.yaml
```diff
@@ -33,6 +33,12 @@ Statement:
       - ec2:StopInstances
       - ec2:TerminateInstances
       - ec2:UnassignPrivateIpAddresses
+      - lightsail:DeleteInstance
+      - lightsail:GetInstance
+      - lightsail:GetInstances
+      - lightsail:RebootInstance
+      - lightsail:StartInstance
+      - lightsail:StopInstance
```
jillr: My apologies @prasadkatti, these actions are actually in the wrong statement - I missed that this block has a condition of ec2:Region: us-east-1 as that line was collapsed by github.  This condition is not met when calling the lightsail API, these need to be moved to another statement.  I believe we should be able to move all of the lightsail actions to a single statement, however I'm getting test errors when testing with 63770 so I haven't fully tested yet.  Moving all these policies to the AllowLightsailCreateInstances statement I can confirm the terminator class appears to run successfully.

prasadkatti: I am sorry, I thought all the tests ran in us-east-1 which is why I added it to that statement. I can move all the lightsail actions to a single statement. 

I still don't fully understand what you mean by -

> This condition is not met when calling the lightsail API

prasadkatti: ohh, I get it! The condition only applies to EC2. :(

jillr: Seems to be, I couldn't find it in the docs but the terminator class (when calling lightsail get_instances) gets an AccessDeniedException when this condition is applied.

prasadkatti: I found the docs. These are the only condition keys defined for lightsail - https://docs.aws.amazon.com/en_pv/IAM/latest/UserGuide/list_amazonlightsail.html#amazonlightsail-policy-keys

The `ec2:Region` condition key is EC2 Service level condition key.

prasadkatti: I have moved all lightsail actions to a single statement. Marking this as resolved.

prasadkatti:
ready_for_review

prasadkatti:
Sounds good. Sorry I didn't read the docs for the Condition key earlier.

jillr:
No worries, I missed it too!

PR #69, opened 2019-10-23
Add Cloudformation permissions

waffie1:
Added permissions to run tests for the ansible module cloudformation_info that are part of pull request https://github.com/ansible/ansible/pull/63008
The tests in that PR need to create/delete a stack as part of the test(which simply creates an SNS topic), as well as various get/describe/list activities that are testing the cloudformation_info module.


waffie1:
i merged something I shouldn't have... closing this and will start clean

PR #70, opened 2019-10-24
Add Permissions for Cloudformation

waffie1:
Add permissions needed to run integration test for ansible cloudformation modules
Add Cloudformation terminator class since this is the first Cloudformation permissions

Added permissions to run tests for the ansible module cloudformation_info that are part of pull request https://github.com/ansible/ansible/pull/63008
The tests in that PR need to create/delete a stack as part of the test(which simply creates an SNS topic), as well as various get/describe/list activities that are testing the cloudformation_info module.

### DIFF
--- a/aws/policy/application-services.yaml
+++ b/aws/policy/application-services.yaml
@@ -93,6 +93,7 @@ Statement:
       - dynamodb:DescribeTable
       - kinesis:ListStreams
       - kinesis:DescribeStream
+      - cloudformation:ListStacks
     Resource: "*"
   - Sid: AllowGlobalResourceRestrictedActionsWhichIncurNoFees
     Effect: Allow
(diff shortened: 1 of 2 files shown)


aws/policy/application-services.yaml
```diff
@@ -180,3 +180,18 @@ Statement:
...
+      - cloudformation:GetStackPolicy
+      - cloudformation:DescribeStackEvents
+      - cloudformation:CreateStack
+      - cloudformation:GetTemplate
+      - cloudformation:DeleteStack
+      - cloudformation:UpdateStack
+      - cloudformation:DescribeChangeSet
+      - cloudformation:ListChangeSets
+      - cloudformation:ListStackResources
+    Resource: arn:aws:cloudformation:*:*:stack/*/*
```
jillr:
```suggestion
    Resource: arn:aws:cloudformation:us-east-1:{{ aws_account_id }}:stack/ansible-test*
```
We generally keep CI policy restricted to us-east-1, with our account, and prefer to use a predictable resource name in tests.

You'll also need to add `cloudformation:ListStacks` to the`AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees` Sid, to allow the terminator to list stacks.  The rest of these permissions look good.

waffie1: The requested change have been made

PR #76, opened 2019-11-14
Add action to create keypair in lightsail

prasadkatti:
Based on https://github.com/ansible/ansible/pull/63770#issuecomment-553078130, we want to manage the keypair used while creating a lightsail instance within the testsuite itself. This means having the ability to create a keypair in lightsail. This adds the action required for that to the relevant policy statement.

Ansible PR - https://github.com/ansible/ansible/pull/63770


prasadkatti:
ready_for_review

This adds actions to create/delete key pairs. Also adds a terminator class to cleanup stale key pairs in lightsail.

PR #78, opened 2019-11-17
Added lambda:GetFunctionConfiguration permission needed for lambda_info…

stefanhorning:
…module integration tests. See PR https://github.com/ansible/ansible/pull/64548

### DIFF
--- a/aws/policy/compute.yaml
+++ b/aws/policy/compute.yaml
@@ -136,6 +136,7 @@ Statement:
       - lambda:CreateFunction
       - lambda:DeleteFunction
       - lambda:GetFunction
+      - lambda:GetFunctionConfiguration
       - lambda:GetPolicy
       - lambda:RemovePermission
       - lambda:UpdateFunctionConfiguration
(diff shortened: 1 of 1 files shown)


stefanhorning:
added `List`ing permissions now as per discussion in https://github.com/ansible/ansible/pull/64548#pullrequestreview-315092589

aws/policy/compute.yaml
```diff
@@ -136,7 +136,11 @@ Statement:
       - lambda:CreateFunction
       - lambda:DeleteFunction
       - lambda:GetFunction
+      - lambda:GetFunctionConfiguration
       - lambda:GetPolicy
+      - lambda:ListVersionsByFunction
+      - lambda:ListAliases
+      - lambda:ListEventSourceMappings
```
jillr: The List* actions need to be added at line 62, where they would be applied to `Resource: "*"`. This block only affects `arn:aws:lambda:us-east-1:{{ aws_account_id }}:function:*` which is insufficient for the list functions.

I have tested 64548 against a policy that has these list function in that section, and everything runs with that change.

stefanhorning: Done

PR #79, opened 2019-11-19
Allow the use of 'shippable-*' as a role name

tremble:
1. Remove duplicate wildcard
2. Allow the use of shippable-* as a role name, this is the value of ``{{ resource_prefix }}``

### DIFF
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -13,18 +13,18 @@ Statement:
       - iam:AddRoleToInstanceProfile
       - iam:RemoveRoleFromInstanceProfile
     Resource:
-      - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-sts-*
-      - arn:aws:iam::{{ aws_account_id }}:instance-profile/ansible-test-sts-*
-      - arn:aws:iam::{{ aws_account_id }}:instance-profile/ansible-test-*
       - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-*
+      - arn:aws:iam::{{ aws_account_id }}:role/shippable-*
+      - arn:aws:iam::{{ aws_account_id }}:instance-profile/ansible-test-*
+      - arn:aws:iam::{{ aws_account_id }}:instance-profile/shippable-*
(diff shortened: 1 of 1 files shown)


mattclay:
@tremble What's the purpose of this change?

tremble:
@mattclay The vast bulk of our AWS test resources are named "{{ resource_prefix }}-something-or-other".  Running 'locally' this shows up as "ansible-test-......-something-or-other".  However, when we run the tests through shippable they show up as "shippable-......-something-or-other".

The exception here is roles, where we're doing weird things like "ansible-test-sts-{{ resource_prefix }}", because that's what this policy had allowed.  I'd like to be more consistent with the tests I'm writing, hence allowing "shippable-*" so I can use resource_prefix as a prefix when writing tests with roles.

mattclay:
We should come up with a way to use a common prefix in both cases, rather than having two different prefixes. This will also avoid making our resource prefix specific to Shippable, which will avoid problems later if tests are run on another CI provider.

Perhaps it's as simple as making the Shippable prefix `ansible-test-shippable`, which has the advantage of more closely matching the existing policy.

jillr:
To recap a part of the irc convo relevant to this PR, one challenge here is that `ansible-test-shippable` results in a `resource_prefix` that is sufficiently long enough to cause some tests to fail on some value exceeding the allowable length (role name or elb name) as seen in https://app.shippable.com/github/ansible/ansible/runs/151499/116/tests.
Related: https://github.com/ansible/ansible/pull/61806

Generally though I'm +1 finding a workable, common prefix.

mattclay:
I'm OK with using a shorter prefix, such as `at-` for the tests. However, we should still support the old prefix in the policy so we don't break tests in other branches. We could update the tests in the other branches, but it would still break tests from older commits, such as when bisecting an issue. That's why I originally suggested using the existing longer prefix.

tremble:
@jillr @mattclay : Over on ansible/ansible#61806 I've switched things out to use "ansible-test-shippable-...." (or just "ansible-test-...." for some of the resources with stricter limits) and made the adjustments required to avoid the resource name character limits.

Just for cleanliness sake I'd still suggest pruning out ```ansible-test-sts-*```

PR #86, opened 2020-01-14
redshift terminator - filter out clusters that are being created

s-hertel:
Initializing redshift clusters can take a couple minutes. Until the ClusterStatus != 'creating', there are keys missing from describe_clusters, including 'ClusterCreateTime'. `describe_clusters` doesn't take a filter so I added filtering after the query.


s-hertel:
Deployed the new terminator lambda.

PR #87, opened 2020-02-13
add ListExports Action

mmoyle:
Add action for module integration test

### DIFF
--- a/aws/policy/application-services.yaml
+++ b/aws/policy/application-services.yaml
@@ -203,6 +203,7 @@ Statement:
       - cloudformation:GetStackPolicy
       - cloudformation:GetTemplate
       - cloudformation:ListChangeSets
+      - cloudformation:ListExports
       - cloudformation:ListStackResources
       - cloudformation:UpdateStack
       - cloudformation:UpdateTerminationProtection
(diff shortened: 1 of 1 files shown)


jillr:
Looks like `cloudformation:ListExports` does not act on resource :stack/* but more globally, so this action needs to be at line 97 (as part of `AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees`).
https://docs.aws.amazon.com/IAM/latest/UserGuide/list_awscloudformation.html

If I make that change and test ansible/ansible/pull/67349 in our policy dev environment I turn up an unrelated bug in the tests (will comment on that PR) but `TASK [cloudformation_exports : Read from Exports]` passes without any policy errors.

mmoyle:
> Looks like `cloudformation:ListExports` does not act on resource :stack/* but more globally, so this action needs to be at line 97 (as part of `AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees`).
> https://docs.aws.amazon.com/IAM/latest/UserGuide/list_awscloudformation.html
> 
> If I make that change and test [ansible/ansible/pull/67349](https://github.com/ansible/ansible/pull/67349) in our policy dev environment I turn up an unrelated bug in the tests (will comment on that PR) but `TASK [cloudformation_exports : Read from Exports]` passes without any policy errors.

Thanks, I understand better how to place actions now. I made the change.

PR #89, opened 2020-02-27
Switch ACM to DbTerminator

jillr:
When an invalid cert exists in ACM, describe_certificates (to obtain the
timestamp) fails. They can still be deleted by ARN however.
See: https://github.com/ansible/ansible/issues/67788

### DIFF
--- a/aws/terminator/security_services.py
+++ b/aws/terminator/security_services.py
@@ -322,6 +322,5 @@ class ACMCertificate(DbTerminator):
     def name(self):
         return self.instance['CertificateArn']
 
-
     def terminate(self):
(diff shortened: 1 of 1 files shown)


aws/terminator/security_services.py
```diff
@@ -303,35 +303,25 @@ def terminate(self):
...
 
     @property
     def id(self):
         return self.instance['CertificateArn']
 
     @property
     def name(self):
-        return self.instance['DomainName']
+        return self.instance['CertificateArn']
+
```
mattclay: CI failure is due to the extra blank line here.

PR #93, opened 2020-04-22
Additional IAM permissions for testing Kinesis streams

tremble:
SSIA

See also https://github.com/ansible-collections/community.aws/pull/42

Moved some of the Kinesis permissions into into the "costs incurred" policy (you're charged by the Shard-hour and this cost is increased if you increase the retention policy)


hacking/aws_config/testing_policies/compute-policy.json
```diff
@@ -276,5 +276,23 @@
...
+        {
+            "Sid": "AllowKinesis",
+            "Effect": "Allow",
+            "Action": [
+                "kinesis:AddTagsToStream",
+                "kinesis:CreateStream",
+                "kinesis:DeleteStream",
+                "kinesis:DecreaseStreamRetentionPeriod",
+                "kinesis:DescribeStream",
+                "kinesis:IncreaseStreamRetentionPeriod",
```
jillr, line 288: Looks like we're missing `kinesis:ListStreams` in the hacking policies.

tremble: There's currently no _info module, so nothing uses it.  I'll get it added.

jillr:
The permissions seem fine, deployed to staging and tested with the kinesis PR.  However I had to strip several other lines out of application-services.yaml to do that - with this PR we would exceed the 6144 character max for managed policies.  So I'm going to work on a bit of refactoring before we can merge this one.   machine-learning.yaml and big-data.yaml are both fairly small, if we can condense them somewhere we should be able to break out some things from the larger policies like application-services.yaml and compute.yaml.

PR #95, opened 2020-05-04
adding AWS Glue Job actions to test policies

cdschneider:
per @tremble comments on PR ansible-collections/community.aws#57

Adding `glue:` permissions to test policies in order to support new `aws_glue_job` module coverage in integration tests.


aws/policy/application-services.yaml
```diff
@@ -134,6 +134,13 @@ Statement:
       - elasticfilesystem:DeleteMountTarget
       - kinesis:CreateStream
       - kinesis:DeleteStream
+      - glue:CreateJob
```
jillr, line 137: This action incurs fees so should be under an appropriate Sid.

aws/policy/application-services.yaml
```diff
@@ -134,6 +134,13 @@ Statement:
...
+      - glue:DeleteJob
+      - glue:UpdateJob
+      - glue:GetJob
+      - glue:GetTags
+      - glue:TagResource
```
jillr, line 142: Tag actions operate on `arn:aws:glue:us-east-1:{{ aws_account_id }}:*` (https://docs.aws.amazon.com/glue/latest/webapi/API_GetTags.html) could you please move the tag-related actions into `AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees`?

aws/policy/application-services.yaml
cdschneider: Sorry about that, I didn't realize this! I was following the pricing description [here](https://aws.amazon.com/glue/pricing/), and took this statement as the `glue:CreateJob` action would not incur fees. 

> With AWS Glue, you only pay for the time your ETL job takes to run

I will move this to the correct Sid. Do you also want me to put these policy changes in `big-data.yml` instead of `application-services.yml`? I see there are another few `glue:*` action permissions in there.

jillr:
@cdschneider The changes you have here are good, however the merge conflict is a bit hairy, partially because of #97 and partially I think from the merge commit.  If you don't mind me pushing to your branch I can try to get those sorted out in the next few days.

cdschneider:
I don't mind at all! @jillr

tremble:
@jillr Merge looks to have gone a little wonky (@cdschneider sorry, #97 was pretty major surgery, but should massively simplify updating the policies)

cdschneider:
If it's easier, I'm happy to close this and re-open a new PR with my changes starting with a fresh branch @jillr @tremble

jillr:
@cdschneider Sorry for the late reply, however you prefer to resolve the merge conflict works for me.

jillr (commented):
Hi @cdschneider, the policies look good but we'll also need a new class added to the terminator lambda for Glue Jobs.  The terminator is a lambda function that runs in our CI account and deletes resources that have been online for longer than a set time, in case a CI job fails to clean up after itself for some reason.

You can see an example of the class for cleaning up glue connections here:
https://github.com/mattclay/aws-terminator/blob/master/aws/terminator/data_services.py#L54-L72

cdschneider:
OK @jillr I think that I've addressed all your previous comments, fixed outstanding merge conflicts, and added the `GlueJob` class to the `data_services.py` to be used by the terminator lambda.  Is there anything else outstanding on this PR?

cdschneider:
@jillr Can you please let me know if this PR is still needed, or if #137 is a replacement for this work? Sorry for my slow-responses on addressing the changes you requested, but all requested changes should all be addressed now in this PR (if it is still needed).

tremble:
Looking at the policies added in this PR the Glue actions look they were added in #137 and we should be able to close this PR out.

jillr:
I think tremble is right, community.aws #57 needs a rebase because we no longer use shippable (JFrog discontinued the service) and the PR needs to pick up the new CI config from `main` but I believe we have the policies we need.

PR #97, opened 2020-05-06
Rework the way 'hacking' policies work

tremble:
- Add the standard CI policies to a group
- Add the *extra* 'unsupported test' policies to a second group
- Move 'unsupported test' policies to YAML (matching the CI policies)
- prune out duplicate policies

This should make it easier for folks to predict the behaviour of CI when running tests in their own account.


jillr:
CI was using an outdated container image; updating that has resolved the key failures.

aws/policy/security-services.yaml
```diff
@@ -116,15 +139,15 @@ Statement:
...
   - Sid: AllowPassRole
     Effect: Allow
     Action:
       - iam:PassRole
     Resource:
-      - arn:aws:iam::{{ aws_account_id }}:role/ansible_lambda_role
-      - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-sts-*
-      - arn:aws:iam::{{ aws_account_id }}:role/ansible-test-*
+      - 'arn:aws:iam::{{ aws_account_id }}:role/ansible-test-*'
+      # Legacy - We need to backport ansible-collections/community.aws/63
```
jillr, line 149: I'm less sure that one will be accepted as a backport.  I'm not opposed to you proposing it; I'd +1 it; but it will be ultimately up to the RM.

aws/policy/security-services.yaml
```diff
@@ -4,83 +4,106 @@ Statement:
...
+      - waf:GetRegexMatchSet
+      - waf:GetRegexPatternSet
       - waf:GetRule
-      - waf:CreateRule
+      - waf:GetRuleGroup
+      - waf:GetSizeConstraintSet
+      - waf:GetSqlInjectionMatchSet
+      - waf:GetWebACL
+      - waf:GetXssMatchSet
+      - waf:ListActivatedRulesInRuleGroup
```
jillr, line 79: The only waf Action this list doesn't include is `ListLoggingConfigurations`. In light of the ongoing struggle with 6144 character policy limits we need to look at making more things use wildcards for low-risk actions, like List and Describe.  I'd prefer to keep the List* here.

hacking/aws_config/setup-iam.yml
```diff
@@ -1,55 +1,97 @@
...
+# Creates IAM policies and associates them with two iam_groups. These groups
+# can then be associated with an appropriate user:
 #
-# You can pass -e profile=boto_profile_name if you have a profile that
-# you can use, otherwise use normal AWS methods (env variables, instance
-# profile, etc)
+# Groups:
+# - ansible-integration-ci
+#   - Mirrors the permissions used by the the AWS collections CI
+# - ansible-integrayion-unsupported
```
jillr, line 9:
```suggestion
# - ansible-integration-unsupported
```

hacking/aws_config/setup-iam.yml
```diff
@@ -1,55 +1,97 @@
...
-# works with ansible-test), pass -e region=us-east-2
+# Policies:
+# - AnsibleIntegration-CI-*
+#   - Managed Policies used by the {{ iam_group_prefix }}-ci group
+# - AnsibleIntegration-Unsupported-*
+#   - Managed Policies used by the {{ iam_group_prefix }}-unsupported group
 #
-# Requires 2.4 for iam_managed_policy and iam_group
+# Supported variables:
+# - region: The AWS region that many of the policies will be restricted to
```
jillr, line 20: I'd like to be clearer here than "many".  Are there policies that are hard coded to some region, or could this be said as "The AWS region that any region-restricted policies will be limited to"?

hacking/aws_config/setup-iam.yml
```diff
@@ -1,55 +1,97 @@
...
-        msg: "You must set the iam_group variable"
-      when: iam_group is not defined
 
     - name: Get aws account ID
       aws_caller_info:
-        profile: "{{ profile|default(omit) }}"
       register: aws_caller_info
 
-    - name: Set aws_account_fact
+    - name: Set aws_account fact
```
jillr, line 54:
```suggestion
    - name: Set aws_account_id fact
```

aws/policy/security-services.yaml
tremble: It's a CI only change, depending on when we stop testing against Ansible 2.9 it's probably not worth the effort.  Updated the comment

hacking/aws_config/test_policies/container-policy.yaml
```diff
@@ -0,0 +1,39 @@
...
+    Effect: Allow
+    Action:
+      - ecs:CreateCluster
+      - ecs:DeleteCluster
+      - ecs:RunTask
+      - ecs:StartTask
+      - ecs:StopTask
+    Resource: "*"
+  #- Sid: AllowGlobalResourceRestrictedActionsWhichIncurNoFees
+  #  Effect: Allow
```
jillr, line 29: What's the purpose of these commented out unrestricted policies, here and in the devops, networking, and storage policies below?

tremble: I'm hoping that if we leave them there commented out, then should someone need/want to add some we'll get a consistent naming of the policies.

hacking/aws_config/test_policies/compute.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - elasticloadbalancing:Create*
+      - elasticloadbalancing:Delete*
+      - elasticloadbalancing:DeregisterInstancesFromLoadBalancer
+      - elasticloadbalancing:Describe*
+      - elasticloadbalancing:Modify*
+      - elasticloadbalancing:Register*
+      - elasticloadbalancing:Deregister*
+      - elasticloadbalancing:Remove*
+    Resource:
+      - 'arn:aws:autoscaling:{{ aws_region }}:{{ aws_account_id }}:*'
```
s-hertel, line 67: I think I've tried this and application/network load balancers don't support resource restrictions. Do you know if that's changed?

hacking/aws_config/test_policies/compute.yaml
```diff
@@ -0,0 +1,78 @@
...
+    Resource: "*"
+
+  - Sid: AllowGlobalUnrestrictedResourceActionsWhichFees
+    Effect: Allow
+    Action:
+      - ec2:*LaunchTemplate
+      - ec2:*LaunchTemplateVersion
+      - ec2:*LaunchTemplateVersions
+      - ec2:ReportInstanceStatus
+      - SNS:ListSubscriptions
```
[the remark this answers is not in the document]
tremble: They are in aws/policy/application-services.yaml

One of the key changes here is that we now automatically add the 'CI' policies *and* the 'test' policies, which means that (at least in theory) test_policies only needs to include the extra policies required by unsupported tests

hacking/aws_config/test_policies/compute.yaml
tremble: Nice catch, Describe actions specifically don't support the restrictions, but the others *should*

hacking/aws_config/testing_policies/security-policy.json
```diff
@@ -1,228 +0,0 @@
...
-    "Statement": [
-        {
-            "Action": [
-                "iam:GetGroup",
-                "iam:GetInstanceProfile",
-                "iam:CreateInstanceProfile",
-                "iam:GetPolicy",
-                "iam:GetPolicyVersion",
-                "iam:GetRole",
-                "iam:GetRolePolicy",
```
s-hertel, line 12: I don't see this one anywhere now.

hacking/aws_config/testing_policies/database-policy.json
```diff
@@ -1,123 +0,0 @@
...
-        },
-        {
-          "Sid": "AllowRedshiftManagment",
-          "Action": [
-              "redshift:CreateCluster",
-              "redshift:CreateTags",
-              "redshift:DeleteCluster",
-              "redshift:DeleteTags",
-              "redshift:DescribeClusters",
-              "redshift:DescribeTags",
```
s-hertel, line 93: I don't see this one anymore.

aws/policy/networking.yaml
```diff
@@ -48,6 +48,7 @@ Statement:
       - ec2:DeleteRoute
       - ec2:AssociateRouteTable
       - ec2:DisassociateRouteTable
+      - ec2:ReplaceRouteTableAssociation
```
s-hertel, line 51: I don't think we actually have a module that does this.

tremble: Yeah, it was in the 'hacking' policy and since we allow associate and disassociate it seemed reasonable to move it into the main list, I can remove it if you'd rather.

s-hertel: Yeah, I saw why you did it. I'd personally prefer to keep the CI policies minimal.

PR #100, opened 2020-05-15
Refactor some things out of application-services.yaml. Again.

jillr:
Application services policies exceeded 6144 chars. Again. Already.


jillr:
(Next time I'm going to just make a new policy file; application-services-2-electric-boogaloo.yml)

PR #102, opened 2020-06-04
Refactor policies again

jillr:
Application services policy is oversize again, this needs to be a more drastic change.


jillr:
I wanted to refactor more things in application services but this is holding up other prs now.

PR #104, opened 2020-07-15
Add tests against 'hacking' policies and add a size test

tremble:
- Add standard linting tests for hacking policies
- Remove defunct script
- Add test for policy size


hacking/aws_config/test-policies.yml
```diff
@@ -0,0 +1,60 @@
...
+    - name: 'Fail if policy is too large'
+      vars:
+        _query: 'results[?changed].item'
+        oversized_policies: "{{ policy_size | json_query(_query) | list }}"
+      debug:
+        var: oversized_policies
+      failed_when:
+        - policy_size is changed
+
+    - name: Create location for policies
```
tremble, line 36: This bit was me starting to play with using AWS tooling to validate the policies, so I needed a compiled copy.  I'm good with this just being removed, when I commented it out it generated extra linting warnings, so I just left it there in case you had a way to test the policies.

jillr: What kind of tooling?  Policy Validator?  It might be interesting to provide something to do that, but obviously we don't have a module for that now.

hacking/aws_config/test-policies.yml
```diff
@@ -0,0 +1,60 @@
...
+
+  tasks:
+    - name: 'Check size of policies'
+      with_fileglob:
+        - '{{ test_policies }}'
+        - '{{ ci_policies }}'
+      set_fact:
+        policy: "{{ lookup('template', item) | from_yaml | to_json }}"
+      changed_when:
+        - policy_size | length > 9216
```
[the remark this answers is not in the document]
tremble: The technical limit is 10240, I used 9216 to give us a little headroom, but maybe 10000 would be a better value?

hacking/aws_config/test-policies.yml
```diff
@@ -0,0 +1,60 @@
...
+    test_policies: 'test_policies/*.yaml'
+
+  tasks:
+    - name: 'Check size of policies'
+      with_fileglob:
+        - '{{ test_policies }}'
+        - '{{ ci_policies }}'
+      set_fact:
+        policy: "{{ lookup('template', item) | from_yaml | to_json }}"
+      changed_when:
```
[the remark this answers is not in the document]
tremble: I was trying to ensure that all failures could be identified at once rather than run, fix, run, fix.  Assert and failed both result in an instant failure when just one fails.  When I tried ignore_errors it didn't try the other entries in the list.

hacking/aws_config/test-policies.yml
tremble: AWS CLI allows you to upload a policy to their simulation tool.  It's not as fussy (or wasn't last I checked) as IAM itself, but it's better than nothing.

Same idea that I was using in https://github.com/ansible/ansible/pull/67581

jillr (approved):
Thanks @tremble - apologies for how long it took to get this one in!

PR #109, opened 2020-09-02
Extend storage services permissions for s3 bucket metrics configuration 

vorotech:
Required for new s3_metrics_configuration module
* `s3:GetMetricsConfiguration` covers get and list s3 bucket metrics configuration operations
* `s3:PutMetricsConfiguration` covers put and delete s3 bucket metrics configuration operations

See https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketMetricsConfiguration.html

### DIFF
--- a/aws/policy/storage-services.yaml
+++ b/aws/policy/storage-services.yaml
@@ -32,4 +32,6 @@ Statement:
       - s3:PutBucketNotification
       - s3:GetBucketLogging
       - s3:PutBucketLogging
+      - s3:GetMetricsConfiguration
+      - s3:PutMetricsConfiguration
     Resource: "*"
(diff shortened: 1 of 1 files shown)


aws/policy/storage-services.yaml
```diff
@@ -32,4 +32,6 @@ Statement:
       - s3:PutBucketNotification
       - s3:GetBucketLogging
       - s3:PutBucketLogging
+      - s3:GetMetricsConfiguration
+      - s3:PutMetricsConfiguration
```
jillr, line 36: Request metrics are billed at CloudWatch pricing so the Put will need to be under a new AllowGlobalUnrestrictedResourceActionsWhichIncurFees Sid.  Otherwise looks good - thanks!

jillr (approved):
LGTM, tested on dev with https://github.com/ansible-collections/community.aws/pull/217 - thanks!

PR #110, opened 2020-09-10
Permit Amazon Autoscaling to automatically create it's default ServiceLinkedRoles

tremble:
When you first try to use an AWS service that makes use of ServiceLinkedRoles it will automatically try to create a Role behind the scenes for you (using your permissions).  This role has very limited access and neither the permissions nor the assume_policy can be changed.

https://aws.amazon.com/blogs/security/introducing-an-easier-way-to-delegate-permissions-to-aws-services-service-linked-roles/


jillr (approved):
LGTM, tests in the linked PR pass.

PR #114, opened 2020-10-23
Add permissions for Lightsail static IPs

danielcotton:
For integration tests involving the querying and creation/deletion of Lightsail Static IPs

Per https://github.com/ansible-collections/community.aws/pull/259 I am trying to add the ability to allocate and release Ligthsail Static IPs from within Ansible. These permissions are necessary for integration tests to pass.


danielcotton:
I'd appreciate it if this could be merged so work can proceed on expanding lightsail functionality in Ansible

jillr (commented):
Thanks @danielcotton. The permissions look good when tetsing with your PR but we'll need a new terminator class (like this one: https://github.com/mattclay/aws-terminator/blob/master/aws/terminator/compute.py#L378-L394) to remove static IPs in the event that the test fails for some reason before the `always` block runs.

danielcotton:
Thank you @jillr. I believe I've made the necessary changes.

jillr (approved):
LGTM, thanks @danielcotton

PR #115, opened 2020-11-13
Replace deprecated service-role policy

tremble:
Thanks to @maxeaubrey (https://github.com/ansible-collections/community.aws/pull/278) the failures for the aws_ssm connection plugin have been identified as related to Amazon deprecating a policy: https://aws.amazon.com/blogs/mt/applying-managed-instance-policy-best-practices/

```
Plan for replacing the AmazonEC2RoleforSSM policy

To help you through the transition of your existing environments, the AmazonEC2RoleforSSM policy remains in place after the introduction of the new AmazonSSMManagedInstanceCore policy.

In the near future, the AmazonEC2RoleforSSM policy will be deprecated. The policy will continue to provide the included permissions to any currently attached users, groups, and roles. However, it will not be available for attachment to new resources and cannot be re-attached once detached from a resource. For more details, see Deprecated AWS Managed Policies.
```

Testing against a separate account I have been able to confirm that using AmazonSSMManagedInstanceCore instead of AmazonEC2RoleforSSM gets the tests running.


jillr:
@tremble the policy change itself looks fine but I'm getting attachment errors when trying to run the tests against staging  
```
Unable to attach policy arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM to role ansible-test-ansible-test-lab-62411740-aws-ssm-role: An error occurred (AccessDenied) when calling the AttachRolePolicy operation: User: arn:aws:sts::966509639900:assumed-role/ansible-core-ci-test-dev/dev=remote=jillr  is not authorized to perform: iam:AttachRolePolicy on resource: role ansible-test-ansible-test-lab-62411740-aws-ssm-role"
```

tremble:
@jillr - We also need https://github.com/ansible-collections/community.aws/pull/295/commits/5163d2076a51c47a4a1913664bbfc5a6b94d89b6

Which switches the policy it uses.

jillr (approved):
I get token errors when testing with community.aws/pull/295, but I believe that's unrelated to this change and we still need to deal with the policy deprecation. This lgtm and I'll leave a comment on the other PR.

PR #116, opened 2020-11-14
Increase age_limit on ec2 dependencies

jillr:
We're getting terminator failures when a resource is still attached to an EC2 instance that is still in the process of terminating. Let these age an extra 5min to reduce log noise.


jillr:
@samdoran I don't think this catches all of them but should help reduce the noise

samdoran:
I went ahead and deployed this change. Thanks for the fix @jillr!

PR #122, opened 2021-01-27
New actions in storage-services - PR#260 ansible-collections/community.aws

zeten30:
Integration tests were extended and hence they need more privileges.

https://github.com/ansible-collections/community.aws/pull/260


jillr (approved):
lgtm - deployed into our CI account. Thanks @zeten30

PR #125, opened 2021-02-17
Add IAM SAML Provider support to terminator and CI

tremble:


jillr (approved):
lgtm, tested with 419

PR #126, opened 2021-02-17
Allow a bunch of elasticache actions for CI testing elasticache module

stefanhorning:
Needed for PR https://github.com/ansible-collections/community.aws/pull/410

### DIFF
--- a/aws/policy/data-services.yaml
+++ b/aws/policy/data-services.yaml
@@ -30,6 +30,21 @@ Statement:
       - dynamodb:DeleteTable
       - dynamodb:UpdateTable
       - dynamodb:TagResource
+      - elasticache:AddTagsToResource
+      - elasticache:CreateCacheSecurityGroup
+      - elasticache:CreateCacheSubnetGroup
+      - elasticache:DeleteCacheCluster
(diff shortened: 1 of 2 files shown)


markuman (approved):
LGTM

jillr (changes requested):
Hi @stefanhorning, the policies here look good but we'll also need a new terminator class for Elasicache since this is a new service being enabled in CI.  There are some examples here: https://github.com/mattclay/aws-terminator/blob/master/aws/terminator/data_services.py

stefanhorning:
@jillr Done now. I hope I guessed the purpose of the methods correctly from other examples.

jillr (approved):
There's a few more permissions needed here, and it turns out CacheClusterCreateTime isn't available until CacheClusterStatus is available so the terminator class will need to ignore clusters with a pending state.  I've got a diff that works with 410 so I'll merge that after this one so you won't be held up any longer.

I didn't fully review your patch @stefanhorning but I'll leave some comments from my testing over there.  Thanks for your work on this!

stefanhorning:
Thanks for pushing this forward!

PR #129, opened 2021-02-21
Support passing a configuration file

tremble:
I'm using aws-terminator to test changes and to cleanup after non-CI tests.  I'd like to be able to keep my config somewhere else.


jillr (approved):
tested locally, lgtm - thanks tremble!

PR #137, opened 2021-03-16
Add AWS Glue Job permissions

ichekaldin:
This allows AWS Glue integration tests to be executed, per @tremble
recommendation in https://github.com/ansible-collections/community.aws/pull/480

This replaces https://github.com/mattclay/aws-terminator/pull/95 which appears to have been abandoned.

### DIFF
--- a/aws/policy/application-services.yaml
+++ b/aws/policy/application-services.yaml
@@ -103,6 +103,12 @@ Statement:
       - kinesis:RemoveTagsFromStream
       - kinesis:StartStreamEncryption
       - kinesis:StopStreamEncryption
+      - glue:DeleteJob
+      - glue:GetJob
+      - glue:GetTags
+      - glue:TagResource
+      - glue:UntagResource
+      - glue:UpdateJob
     Resource:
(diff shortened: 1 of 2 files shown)


jillr (commented):
Hi @ichekaldin, the policies look good but we'll also need a new class added to the terminator lambda for Glue Jobs. The terminator is a lambda function that runs in our CI account and deletes resources that have been online for longer than a set time, in case a CI job fails to clean up after itself for some reason.

You can see an example of the class for cleaning up glue connections here:
https://github.com/mattclay/aws-terminator/blob/master/aws/terminator/data_services.py#L54-L72

ichekaldin:
@jillr I added the class to cleanup Glue jobs. Thank you!

ichekaldin:
@jillr Is there anything else that I'm missing for this PR?

jillr:
Hi @ichekaldin, apologies for the delay I was out of the office for a bit - I work through this queue and do policy deploys on Thursdays usually.  These seem good but I'm not able to test successfully with 480, I'll leave some comments over there.

jillr (approved):

PR #138, opened 2021-03-21
Move SNS over into 'supported' policy

tremble:
We already have the terminator class in place, add the permissions so we can run the SNS tests.


tremble:
This includes permitting **read-only** access to the SNS queue to which Amazon publicly publishes changes in its IP address ranges.  The queue is owned by Amazon and the data is known to be safe.

jillr (approved):

PR #139, opened 2021-03-21
Add support for DMS Endpoints

tremble:
We have initial integration tests written, get them running in CI.


jillr:
I'm fine with these policies, but we're oversize again.

jillr:
I (finally - my apologies) opened a refactor.  In the interest of the git conflict that's going to cause here I snuck these new policies into that PR.  Assuming that lands ok, 150 would solve for this.

PR #144, opened 2021-04-21
Policies for rds_option_group management

alinabuzachis:
Add policies for new modules -  `rds_option_group` and `rds_option_group_info` https://github.com/ansible-collections/community.aws/pull/517


alinabuzachis:
> Thanks @alinabuzachis. The permissions look good and your tests pass (after adding the module names to meta/runtime.yml) but you'll also need to add a terminator class for RdsOptionGroup to aws/terminator/data_services.py. It looks like `describe_option_group` doesn't provide a created_at timestamp, so you'll need to subclass `DbTerminator` (the way `RdsDbParameterGroup` is) instead of `Terminator`. That tells the terminator to store a record of the first time it saw a resource in its own database so it knows when to expire something.
> 
> I also found that RDS creates a default option group for each DB type the first time you make one, that can not be deleted. https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_WorkingWithOptionGroups.html So you'll need to add an `ignore` parameter to the new class for option names that start with `default:`, kind of like how these policies are ignored for IAM:
> https://github.com/mattclay/aws-terminator/blob/master/aws/terminator/security_services.py#L25

Thank you very much for explaining. I tried to add the RdsOptionGroup class.

aws/terminator/data_services.py
```diff
@@ -158,3 +158,24 @@ def created_time(self):
...
+    def id(self):
+        return self.instance['OptionGroupArn']
+
+    @property
+    def name(self):
+        return self.instance['OptionGroupName']
+
+    @property
+    def ignore(self):
+        return not self.name.startswith('default')
```
jillr, line 178:
```suggestion
        return self.name.startswith('default')
```
This one is a little confusing because it's kind of a double negative (I had to experiment with it a bit myself).  This function returns a list of things _to be ignored_, not a list of things that excludes what to ignore.  So we want to return only those things which start with `default`.

alinabuzachis: Ok, done. Thank you.

jillr (approved):
resolved the merge conflict, LGTM

PR #146, opened 2021-04-22
Add S3 object tagging policies

alinabuzachis:
Add S3 object tagging policies required in https://github.com/ansible-collections/amazon.aws/pull/335


jillr (approved):
lgtm, thanks!

PR #147, opened 2021-04-25
Add AWS Glue Crawler permissions

ichekaldin:
This allows AWS Glue crawler integration tests to be executed:
https://github.com/ansible-collections/community.aws/pull/546

### DIFF
--- a/aws/policy/application-services.yaml
+++ b/aws/policy/application-services.yaml
@@ -66,6 +66,9 @@ Statement:
       - kinesis:DescribeStream
       - cloudformation:DescribeStacks
       - cloudformation:ListExports
+      - glue:GetConnections
+      - glue:GetCrawlers
+      - glue:GetJobs
     Resource: "*"
   - Sid: AllowGlobalResourceRestrictedActionsWhichIncurNoFees
     Effect: Allow
--- a/aws/policy/application-services.yaml
+++ b/aws/policy/application-services.yaml
@@ -107,12 +107,15 @@ Statement:
       - kinesis:RemoveTagsFromStream
       - kinesis:StartStreamEncryption
       - kinesis:StopStreamEncryption
+      - glue:DeleteCrawler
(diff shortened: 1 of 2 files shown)


aws/policy/application-services.yaml
```diff
@@ -107,12 +107,15 @@ Statement:
       - kinesis:RemoveTagsFromStream
       - kinesis:StartStreamEncryption
       - kinesis:StopStreamEncryption
+      - glue:DeleteCrawler
       - glue:DeleteJob
+      - glue:GetCrawler
```
jillr, line 112: We also need glue:GetCrawlers, for use by the terminator class.  This needs to operate globally, so it should go under `  - Sid: AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees`

jillr (changes requested):
Thanks @ichekaldin. The permissions and terminator that are here look ok but I'm not able to test them fully with your other PR.  I'll leave some comments on 546.

aws/policy/application-services.yaml
ichekaldin: @jillr Thank you for reviewing this - that's an interesting point.

I see that there are no other Glue-related permissions under `Sid: AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees`.

If they are in fact needed for terminator class, should I add permissions for Glue crawlers, jobs, and connections:

```
- glue:GetConnections
- glue:GetCrawlers
- glue:GetJobs
```

What do you think?

jillr: @ichekaldin Yes please, thanks!

ichekaldin: @jillr Done

jillr (approved):
Thanks @ichekaldin  I'm still getting boto errors running the test suite locally but these policies look good.  I'll get these merged and deployed so we can debug in CI.

PR #148, opened 2021-05-05
Update permissions for community.aws PR 565

badnetmask:
As recommended by @tremble, this PR updates permissions that allows integration tests to run for [community.aws PR 565](https://github.com/ansible-collections/community.aws/pull/565).

### DIFF
--- a/aws/policy/networking.yaml
+++ b/aws/policy/networking.yaml
@@ -40,7 +40,6 @@ Statement:
       - ec2:CreateRouteTable
       - ec2:CreateSecurityGroup
       - ec2:CreateSubnet
-      - ec2:CreateTags
       - ec2:CreateVpc
       - ec2:CreateVpcEndpoint
       - ec2:CreateVpcPeeringConnection
--- a/aws/policy/networking.yaml
+++ b/aws/policy/networking.yaml
@@ -10,6 +10,8 @@ Statement:
       - route53:GetHostedZone
       - route53:DeleteHostedZone
       - route53:UpdateHostedZoneComment
+      - route53:ChangeTagsForResource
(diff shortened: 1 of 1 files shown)


aws/policy/networking.yaml
```diff
@@ -38,6 +40,7 @@ Statement:
       - ec2:CreateRouteTable
       - ec2:CreateSecurityGroup
       - ec2:CreateSubnet
+      - ec2:CreateTags
```
tremble, line 43:
```suggestion
```

aws/policy/networking.yaml
```diff
@@ -71,6 +74,7 @@ Statement:
       - ec2:DescribeRouteTables
       - ec2:DescribeSecurityGroups
       - ec2:DescribeSubnets
+      - ec2:DescribeTags
```
tremble, line 77:
```suggestion
```

The EC2 Tags permissions are over in the compute policy and should already be there.

tremble (changes requested):

aws/policy/networking.yaml
jillr: ec2:DescribeTags here duplicates the ec2:DescribeTags in compute.yaml.  Amazon limits policy size (by character count) and we tend to run into those limits so it's helpful to avoid duplication.

jillr (changes requested):
One small thing, otherwise the route53 permissions look good and the tests from 565 pass for me locally

badnetmask:
> One small thing, otherwise the route53 permissions look good and the tests from 565 pass for me locally

Pushed an update. Thanks.

jillr (approved):
Thanks @badnetmask!

PR #149, opened 2021-05-10
secret policies + terminator class

abikouo:
Required for pull request: https://github.com/ansible-collections/amazon.aws/pull/351


jillr:
Thanks @abikouo!  You have absolutely the right idea here, but we've needed to do larger a refactor for a while since security-services.yaml wasn't the only policy file that needed work.  I've done that in #150.  It did involved a new file, but once I sat down and looked at what things could logically move together I ended up creating an application-security file instead.  Once that lands, I checked and there will be enough space in security-services.yaml for these permissions.

The permissions listed here look good to me.  We'll also need a terminator class though, similar to this https://github.com/mattclay/aws-terminator/blob/master/aws/terminator/security_services.py#L47:L72

abikouo:
> Thanks @abikouo! You have absolutely the right idea here, but we've needed to do larger a refactor for a while since security-services.yaml wasn't the only policy file that needed work. I've done that in #150. It did involved a new file, but once I sat down and looked at what things could logically move together I ended up creating an application-security file instead. Once that lands, I checked and there will be enough space in security-services.yaml for these permissions.
> 
> The permissions listed here look good to me. We'll also need a terminator class though, similar to this https://github.com/mattclay/aws-terminator/blob/master/aws/terminator/security_services.py#L47:L72

@jillr thanks for your feedback on this PR. 
I will wait for #150 to be merged then realign on this and also create the terminator class

jillr:
@abikouo 150 has been merged, thanks!

abikouo:
@jillr could you please review this one?

aws/policy/security-services.yaml
```diff
@@ -95,6 +95,20 @@ Statement:
       - kms:UpdateKeyDescription
     Resource: "*"
 
+  - Sid: AllowGlobalRestrictedResourceActionsWhichIncurFees
+    Effect: Allow
+    Action:
+      - secretsmanager:DescribeSecret
+      - secretsmanager:ListSecrets
```
jillr, line 102: Both Describe and List need to be in the above group (`AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees`, operating on `Resource: "*"`)

aws/policy/security-services.yaml
```diff
@@ -95,6 +95,20 @@ Statement:
...
+      - secretsmanager:CreateSecret
+      - secretsmanager:UpdateSecret
+      - secretsmanager:DeleteSecret
+      - secretsmanager:GetSecretValue
+      - secretsmanager:RotateSecret
+      - secretsmanager:UntagResource
```
jillr, line 108:
```suggestion
      - secretsmanager:UntagResource
      - secretsmanager:TagResource
```

jillr (commented):
With the noted comments I was able to get your tests in #351 to pass, and the terminator class works.  Slack me when you're ready for a review and I'll look asap!  Thanks for your work on this, sorry for the delays.

jillr (approved):
Passing with 351, thanks very much for your work on this!

PR #153, opened 2021-07-02
add DisableMetricsCollection for ec2_asg

goneri:


jillr:
@goneri this one seems fine, it's helpful if you could please link a supporting PR for policy changes - that way we know if they're reasonable permissions to open up and can test the change.

jillr:
@goneri I believe these permissions are already available, re-open this if I missed something.

PR #155, opened 2021-07-09
Added describe event rule permission

avishayil:
I made a PR for Ansible AWS community collection that adds support for `InputTransformer` attribute of `cloudwatchevent_rule`.

This permission should allow the test to pass.

https://github.com/ansible-collections/community.aws/pull/623

PR #158, opened 2021-07-22
Additional cloudwatch_event policies

jillr:
Follow up to #155, these required testing through the commmunity.aws PR locally to uncover so adding them in a second PR


jillr:
going to have to refactor some of the Sids in application-services.yaml  for size limits

PR #162, opened 2021-09-16
Redshift subnet group terminator support

tremble:
https://github.com/ansible-collections/community.aws/pull/719


jillr (commented):
We need ` 'arn:aws:redshift:{{ aws_region }}:{{ aws_account_id }}:subnetgroup:*'` as a resource.

Everything else lgtm as tested with community.aws/pull/719

jillr (approved):
thanks!

PR #164, opened 2021-09-20
Add RDS missing permissions

marknet15:
## Overview
I'm in the progress of making a small RDS change here to fix a bug with enabling enhanced monitoring, and as part of that I need to enable the integration tests.
https://github.com/ansible-collections/community.aws/pull/712

Currently it seems some permissions are missing for example `DescribeDBInstances`:

```
TASK [rds_instance : Ensure the resource doesn't exist] ************************
task path: /home/zuul/.ansible/collections/ansible_collections/community/aws/tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_tags.yml:4
The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_rds_instance_payload_x6h5cji8/ansible_rds_instance_payload.zip/ansible_collections/community/aws/plugins/modules/rds_instance.py", line 811, in get_instance
  File "/home/zuul/venv/lib/python3.6/site-packages/botocore/client.py", line 337, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/zuul/venv/lib/python3.6/site-packages/botocore/client.py", line 656, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the DescribeDBInstances operation: User: arn:aws:sts::966509639900:assumed-role/ansible-core-ci-test-dev/dev=remote=zuul-cloud is not authorized to perform: rds:DescribeDBInstances on resource: arn:aws:rds:us-east-1:966509639900:db:ansible-test-52620875-ip-172-16-123-238-tags
fatal: [tags]: FAILED! => {
    "boto3_version": "1.15.0",
    "botocore_version": "1.18.0",
    "changed": false,
    "error": {
        "code": "AccessDenied",
        "message": "User: arn:aws:sts::966509639900:assumed-role/ansible-core-ci-test-dev/dev=remote=zuul-cloud is not authorized to perform: rds:DescribeDBInstances on resource: arn:aws:rds:us-east-1:966509639900:db:ansible-test-52620875-ip-172-16-123-238-tags",
        "type": "Sender"
    },
```

This also adds specific IAM permissions needed for policy attachments for the enhanced monitoring

### DIFF
--- a/aws/policy/data-services.yaml
+++ b/aws/policy/data-services.yaml
@@ -14,9 +14,7 @@ Statement:
       - glue:DeleteConnection
       - glue:UpdateConnection
       - glue:GetConnections
-      - rds:DescribeDBClusters
-      - rds:DescribeDBParameterGroups
-      - rds:DescribeDBSubnetGroups
+      - rds:DescribeDB*
     Resource: "*"
   - Sid: AllowGlobalResourceRestrictedActionsWhichIncurNoFees
     Effect: Allow
@@ -80,7 +78,6 @@ Statement:
       - rds:DeleteDBCluster
       - rds:DeleteDBParameterGroup
       - rds:DeleteDBSubnetGroup
-      - rds:DescribeDB*
       - rds:RestoreDBInstanceToPointInTime
       - rds:CreateDBInstanceReadReplica
       - rds:CreateDBInstance
@@ -88,6 +85,7 @@ Statement:
       - rds:DeleteDBInstance
       - rds:StopDBInstance
       - rds:StartDBInstance
+      - rds:PromoteReadReplica
       - rds:ListTagsForResource
       - rds:ModifyDBCluster
       - rds:ModifyDBParameterGroup
--- a/aws/policy/data-services.yaml
+++ b/aws/policy/data-services.yaml
@@ -75,11 +75,11 @@ Statement:
       - rds:DeleteDBCluster
       - rds:DeleteDBParameterGroup
       - rds:DeleteDBSubnetGroup
-      - rds:DescribeDBInstances
+      - rds:DescribeDB*
+      - rds:CreateDBInstanceReadReplica
       - rds:CreateDBInstance
       - rds:ModifyDBInstance
       - rds:DeleteDBInstance
-      - rds:DescribeDBParameters
       - rds:ListTagsForResource
       - rds:ModifyDBCluster
       - rds:ModifyDBParameterGroup
@@ -89,6 +89,8 @@ Statement:
       - rds:CreateOptionGroup
       - rds:ModifyOptionGroup
       - rds:DeleteOptionGroup
+      - rds:StopDBInstance
+      - rds:StartDBInstance
     Resource:
       - 'arn:aws:dms:{{ aws_region }}:{{ aws_account_id }}:subgrp:*'
       - 'arn:aws:dynamodb:{{ aws_region }}:{{ aws_account_id }}:table/*'
--- a/aws/policy/data-services.yaml
+++ b/aws/policy/data-services.yaml
@@ -71,6 +71,7 @@ Statement:
       - rds:DeleteDBCluster
       - rds:DeleteDBParameterGroup
       - rds:DeleteDBSubnetGroup
+      - rds:DescribeDBInstances
       - rds:DescribeDBParameters
       - rds:ListTagsForResource
       - rds:ModifyDBCluster
--- a/aws/policy/data-services.yaml
+++ b/aws/policy/data-services.yaml
@@ -76,10 +76,13 @@ Statement:
       - rds:DeleteDBParameterGroup
       - rds:DeleteDBSubnetGroup
       - rds:DescribeDB*
+      - rds:RestoreDBInstanceToPointInTime
       - rds:CreateDBInstanceReadReplica
       - rds:CreateDBInstance
       - rds:ModifyDBInstance
       - rds:DeleteDBInstance
+      - rds:StopDBInstance
+      - rds:StartDBInstance
       - rds:ListTagsForResource
       - rds:ModifyDBCluster
       - rds:ModifyDBParameterGroup
@@ -89,8 +92,6 @@ Statement:
       - rds:CreateOptionGroup
       - rds:ModifyOptionGroup
       - rds:DeleteOptionGroup
-      - rds:StopDBInstance
-      - rds:StartDBInstance
     Resource:
       - 'arn:aws:dms:{{ aws_region }}:{{ aws_account_id }}:subgrp:*'
       - 'arn:aws:dynamodb:{{ aws_region }}:{{ aws_account_id }}:table/*'
--- a/aws/policy/data-services.yaml
+++ b/aws/policy/data-services.yaml
@@ -86,6 +86,7 @@ Statement:
       - rds:StopDBInstance
       - rds:StartDBInstance
       - rds:PromoteReadReplica
+      - rds:RebootDBInstance
       - rds:ListTagsForResource
       - rds:ModifyDBCluster
       - rds:ModifyDBParameterGroup
--- a/aws/policy/data-services.yaml
+++ b/aws/policy/data-services.yaml
@@ -72,6 +72,9 @@ Statement:
       - rds:DeleteDBParameterGroup
       - rds:DeleteDBSubnetGroup
       - rds:DescribeDBInstances
+      - rds:CreateDBInstance
+      - rds:ModifyDBInstance
+      - rds:DeleteDBInstance
       - rds:DescribeDBParameters
       - rds:ListTagsForResource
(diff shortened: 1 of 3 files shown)


jillr:
The rds tests are slow so it'll take a while to test; at first pass the policies look reasonable.  You'll also need to add a class to the terminator in terminator/data_services.py for RdsInstance or some similar name.  The RdsDbParameterGroup and   RdsOptionGroup classes should be good examples to reference.

jillr:
A couple of things I found testing this one.  I'll note the test task that triggered the AccessDenied failure and the policy that's being reported as missing.  This is as far as I got with it today, thanks for your work on this!

rds:RestoreDBInstanceToPointInTime
tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_restore_instance.yml:34

rds:DescribeDBSnapshots
tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_snapshot.yml:4
For Describe policies they can just be a wildcard, like rds:DescribeDB*

rds:CreateDBInstanceReadReplica
tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_read_replica.yml:43

In policy/security-services.yaml you'll need to add an iam:PolicyArn: condition for the task at  tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_enhanced_monitoring.yml:32
"Unable to attach policy arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole to role ansible-test-lab-65106811-enhanced-monitoring-role: An error occurred (AccessDenied) when calling the AttachRolePolicy opera
tion: User: arn:aws:sts::966509639900:assumed-role/ansible-core-ci-test-dev/dev=remote=jillr is not authorized to perform: iam:AttachRolePolicy on resource: role ansible-test-lab-65106811-enhanced-monitoring-role",

rds:StopDBInstance 
tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_states.yml:86

marknet15:
@jillr thanks again for testing this out, I think I've done the class correctly and I included all the extra permissions

aws/policy/security-services.yaml
```diff
@@ -138,6 +138,14 @@ Statement:
       - 'arn:aws:iam::{{ aws_account_id }}:role/ansible-test-*'
       # This is hard coded into DMS...
       - 'arn:aws:iam::{{ aws_account_id }}:role/dms-vpc-role'
+  - Sid: AllowEnhancedMonitoringPolicyAttach
```
jillr, line 141: Apologies @marknet15 I wasn't able to finish testing this today.  I believe you need to add 'arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole' around line 22 instead of this section.  However because the rds test suite takes so long to run I wasn't able to get a successful test run.  I'll keep working on this review.

So far when running it against the resources created by the test suite the terminator class seems good

jillr: Hm, maybe not - just caught this on my latest test run.  Will need to debug more.
`"msg": "Unable to modify DB instance: An error occurred (InvalidParameterValue) when calling the ModifyDBInstance operation: IAM role ARN value is invalid or does not include the required permissions for: ENHANCED_MONITORING"`

marknet15: @jillr No I think you are correct ! I totally missed that section on policies ! I've removed my blob and added the policy to the allowed list. I think thats the only actual missing bit.

marknet15: Actually I think I know whats happening, I spotted something similar internally but I think there's some eventual consistency delay, when creating the IAM role. After a retry it seems to pass once the IAM role has fully provisioned..

marknet15: I'm best sure how to handle it as it seems intermittent, but I removed the IAM deletion at the end of the tests, as it's not cost incurring and wouldn't change.

tremble: @marknet15 - 'terminator' (the python code living in this module) will delete any roles you leave lying around for too long.  If you're having problems with IAM roles being 'eventually consistent', I find it helps to move the iam role creation towards to the start of the tests.  Given how slow the RDS tests are this should give enough time for everything to settle down.

marknet15: @tremble thanks🙏 yep that’s a good suggestion, I moved the IAM creation to the start earlier, so fingers crossed that sorts it out as I think the issue was eventual consistency.

aws/policy/data-services.yaml
```diff
@@ -80,7 +80,14 @@ Statement:
       - rds:DeleteDBCluster
       - rds:DeleteDBParameterGroup
       - rds:DeleteDBSubnetGroup
-      - rds:DescribeDBParameters
+      - rds:DescribeDB*
```
jillr, line 83: We want to move this into the first Sid, `AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees`, otherwise we need to add more resources (tests got hung up on snapshots) and for Describe actions we mostly don't care.  The 3 Describe actions there now could be condensed down to a single wildcard.

aws/policy/data-services.yaml
```diff
@@ -80,7 +80,14 @@ Statement:
...
+      - rds:RestoreDBInstanceToPointInTime
+      - rds:CreateDBInstanceReadReplica
+      - rds:CreateDBInstance
+      - rds:ModifyDBInstance
+      - rds:DeleteDBInstance
+      - rds:StopDBInstance
+      - rds:StartDBInstance
```
jillr, line 90: We also need to add PromoteReadReplica, getting:
```
    "msg": "Unable to promote read replica: An error occurred (AccessDenied) when calling the PromoteReadReplica operation: User: arn:aws:sts::966509639900:assumed-role/ansible-core-ci-test-dev/dev=remote=jillr is not authorized to perform: rds:PromoteReadReplica on resource: arn:aws:rds:us-east-1:966509639900:db:ansible-test-lab-77470753-read-replica-replica",
```

jillr (commented):
Unfortunately the tests are exceeding the 1 hour token duration limit we have on our CI role and are going to have to be either condensed or parallelized.  I was only able to test up to around tests/integration/targets/rds_instance/roles/rds_instance/tasks/test_read_replica.yml:107 before token expiration.  @tremble has done some work getting the [ec2_instance tests](https://github.com/ansible-collections/amazon.aws/tree/main/tests/integration/targets/ec2_instance) to run in parallel but I'm not sure offhand if that approach would work for RDS.

We really appreciate the time you've put into these PRs to get these tests in a better place!

tremble:
> @tremble has done some work getting the [ec2_instance tests](https://github.com/ansible-collections/amazon.aws/tree/main/tests/integration/targets/ec2_instance) to run in parallel but I'm not sure offhand if that approach would work for RDS.

RDS instance tests are already running in parallel.  RDS is just a pretty slow service to work with.

marknet15:
@tremble @jillr No worries :) I'm in the process of condensing the tests right down, as I've spotted some easy ones to do so with.

aws/policy/data-services.yaml
```diff
@@ -80,7 +78,14 @@ Statement:
...
       - rds:DeleteDBSubnetGroup
-      - rds:DescribeDBParameters
+      - rds:RestoreDBInstanceToPointInTime
+      - rds:CreateDBInstanceReadReplica
+      - rds:CreateDBInstance
+      - rds:ModifyDBInstance
+      - rds:DeleteDBInstance
+      - rds:StopDBInstance
+      - rds:StartDBInstance
+      - rds:PromoteReadReplica
```
jillr, line 88:
```suggestion
      - rds:PromoteReadReplica
      - rds:RebootDBInstance
```

aws/terminator/data_services.py
```diff
@@ -178,6 +178,23 @@ def terminate(self):
...
+        return Terminator._create(credentials, RdsDBInstance, 'rds', lambda client: client.describe_db_instances()['DBInstances'])
+
+    @property
+    def id(self):
+        return self.instance['DBInstanceArn']
+
+    @property
+    def name(self):
+        return self.instance['DBInstanceIdentifier']
+
```
jillr, line 193:
```suggestion
    @property
    def age_limit(self):
        return datetime.timedelta(minutes=60)
```
The terminator deletes resources after 20 minutes by default, we can override that with this property though.  The class is definitely working though!  (I killed a few databases in the middle of a test run by forgetting about the 20 minute limit  :))

jillr (commented):
This is getting much closer.  Given how slow RDS is as a service generally it might be difficult to get these under 60 minutes, I can plan to try to look at your test PR more closely next week after AnsibleFest and Contributor Summit are wrapped up.

aws/terminator/data_services.py
marknet15: :D yeah that probably isn't ideal !

PR #182, opened 2021-11-26
ECR GetAuthorizatioinToken as unrestriced action

gen2fish:
ecr:GetAuthorizationToken is an unrestricted action. This will move it to the correct placement.


jillr (approved):
Thanks @gen2fish!

PR #187, opened 2022-01-15
Initial policy statement for `opensearch` service

sebastien-rosset:
Add policy for PR https://github.com/ansible-collections/community.aws/pull/859
Not ready for review yet.


alinabuzachis:
recheck

sebastien-rosset:
@jillr , @alinabuzachis , how can I make progress with this? It's not possible to test without having the right permissions.

alinabuzachis:
> @jillr , @alinabuzachis , how can I make progress with this? It's not possible to test without having the right permissions.

@sebastien-rosset We need to wait for @jillr's suggestion. They generally do the triage on this repo on Thursday. I will try to ask and let you know. Thank you.

jillr:
As long as we can bring the policy under the character limit data-services is the closest IMO.  If not, @alinabuzachis we might need to refactor the policies again.  :)

Describe and List actions can be added as wildcards for any AWS services that do not affect security (ie; IAM or STS).  So some of these can be condensed to `es:Describe*` and `es:List*`.  `es:Get*` can be treated the same.  They also should go under a more general Sid (Kafka is a bad example in this policy), like `AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees`.  `CreateDomain`, `UpdateDomainConfig`, and anything else which can affect the pricing used for opensearch should be under a `AllowGlobalRestrictedResourceActionsWhichIncurFees` Sid and limited to a `domain*` resource.  The actual policy actions included here lgtm - thanks @sebastien-rosset!

jillr:
Since this is the first time we're enabling opensearch, we will also need a class added to the terminator: https://github.com/mattclay/aws-terminator/blob/master/aws/terminator/data_services.py
This is a lamnda that we run to automatically check for and clean-up allowed resources after they exceed an age limit (defaults to 20 minutes) in case the CI job fails to delete a resource for some reason.

sebastien-rosset:
@jillr , is this what you were looking for?

gravesm:
I posted on the associated PR, but just to add here, I don't think we should merge these changes at this time. Our CI constraints would currently prevent us from being able to run a test suite that spans several hours.

PR #191, opened 2022-02-02
Add t2.nano to allowed autoscaling instance types

mandar242:
Added missing instance type ```t2.nano``` for ec2_asg integration tests.

### DIFF
--- a/aws/policy/compute.yaml
+++ b/aws/policy/compute.yaml
@@ -41,6 +41,7 @@ Statement:
         autoscaling:InstanceTypes:
           - t3.nano
           - t3.micro
+          - t2.nano
 
   # Permit RunInstance to access any of the usual objects attached to an
   # instance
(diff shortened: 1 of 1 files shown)


aws/policy/compute.yaml
```diff
@@ -41,6 +41,7 @@ Statement:
         autoscaling:InstanceTypes:
           - t3.nano
           - t3.micro
+          - t2.nano
```
jillr, line 44: Aha! There's 2 things we need to change here actually.  The change you have is good, but also the resource it's acting on is `arn:aws:autoscaling:us-east-1:966509639900:autoScalingGroup:85d03fe6-7fdf-4005-bc5d-06e1d11dc5b7:autoScalingGroupName/ansible-test-37463066-lab-asg`.  In 
https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2autoscaling.html it shows the resource that `UpdateAutoScalingGroup` can be restricted to is `autoScalingGroup*` but we have `autoScalingGroup:*` (notice the wildcard is after a `:` in our policy, which means it won't catch `autoScalingGroupName/*` resources).

Can you please edit the resource to be `'arn:aws:autoscaling:{{ aws_region }}:{{ aws_account_id }}:autoScalingGroup*'`? (with no `:` before the `*`)

PR #192, opened 2022-02-07
feat: add permissions to create and delete SSM documents

blinkseb:
Hello,

I'm adding support for SSM documents when using the Ansible's SSM connection plugin, and there's a need to be able to create and delete SSM documents during CI, hence this PR. You can find the Ansible PR here: https://github.com/ansible-collections/community.aws/pull/876

Extract from the job log:

```
An error occurred (AccessDeniedException) when calling the CreateDocument operation: User: arn:aws:sts::966509639900:assumed-role/ansible-core-ci-test-prod/prod=remote=zuul-cloud is not authorized to perform: ssm:CreateDocument on resource: arn:aws:ssm:us-east-1:966509639900:document/ansible-custom-document because no identity-based policy allows the ssm:CreateDocument action
```

Thanks a lot in advance

### DIFF
--- a/aws/policy/application-services.yaml
+++ b/aws/policy/application-services.yaml
@@ -93,6 +93,9 @@ Statement:
   - Sid: AllowGlobalResourceRestrictedActionsWhichIncurNoFees
     Effect: Allow
     Action:
+      - ssm:CreateDocument
+      - ssm:DeleteDocument
+      - ssm:ListDocuments
       - cloudformation:CreateChangeSet
       - cloudformation:CreateStack
       - cloudformation:DeleteChangeSet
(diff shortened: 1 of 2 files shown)


markuman (approved):

blinkseb:
> @blinkseb Thanks for the PR. Since this is adding permissions to create an SSM document, could you also please add a new SsmDocument terminator class?

Sorry for the late answer! I added the terminator class for `SsmDocument`. I had to use the `DbTerminator` base class because the SSM document only has a creation date, and not a creation timestamp.

gravesm:
@blinkseb Sorry for the delay. Could you rebase this and I'll finish the review.

blinkseb:
@gravesm no worry :) I've rebased on top of master

gravesm:
Thanks, this looks good. One quick change, though. Unless I'm missing some nuance of SSM, you shouldn't need to use the `DbTerminator`. The boto documentation is a little misleading and the `CreatedDate` should actually be the full creation timestamp.

blinkseb:
> Thanks, this looks good. One quick change, though. Unless I'm missing some nuance of SSM, you shouldn't need to use the `DbTerminator`. The boto documentation is a little misleading and the `CreatedDate` should actually be the full creation timestamp.

You are right! I just checked with a SSM document, and `CreateDate` is a timestamp, and not a date. I made the changes to not use `DbTerminator`

markuman:
@blinkseb can you add the `ssm:ListDocuments` permissions and also add the filter?

blinkseb:
Sure, sorry for the delay, I've missed the previous request. According to the docs, it should be the correct way to add a filter, but I haven't tested it. Let me know if something needs to be changed. Thanks!

PR #193, opened 2022-02-10
Refactor ASG policies

jillr:
Remove redundant Sids, move ASG policies that needs to be conditionally restricted by instance type to the same Sid as ec2 policies. Move other ASG policies to existing AllowGlobal polcies.

This should enable the tests in [community.aws#815](https://github.com/ansible-collections/community.aws/pull/815)


gravesm (approved):
I was able to deploy the changes and get that PR to pass in staging.

PR #196, opened 2022-03-02
Update requirements

gravesm:
The recent network-firewall changes require a newer version of botocore. This upgrade triggers a cascade of updates for the other requirements. The lambda runs without errors from a clean virtualenv with these requirements.

### DIFF
--- a/aws/requirements.txt
+++ b/aws/requirements.txt
@@ -1,9 +1,12 @@
-boto3==1.15.18
-botocore==1.18.18  # 1.19.0 (and boto3 1.16.0) requires a urllib3 update
-docutils==0.14
-jmespath==0.9.4
-python-dateutil==2.8.0
-requests==2.22.0
-s3transfer==0.3.0
-six==1.12.0
-urllib3==1.24.3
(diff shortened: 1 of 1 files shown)


gravesm:
cc: @mattclay

jillr (commented):
@gravesm Just to confirm, have you tested these all work in the AWS Laambda Python environment?

gravesm:
From the virtualenv with these dependencies installed, I ran `python cleanup.py --stage dev -c` and it completes with no errors. This is using the new network-firewall changes. I guess I'm a little unclear if there's some other way to test these without actually deploying the changes.

jillr:
You'd have to deploy them to Lambda somewhere.  You should be able to switch what account the app is deployed to in aws/config.yml and your account on the content team AWS account should have the policies necessary to use Lambda.

gravesm:
Just confirming it does work in the Lambda environment.

jillr (approved):

jillr:
Thanks, merge and deploy at will!

PR #197, opened 2022-03-03
Add aws_eks_fargate_profile permissions 

tjarra:
Permissions to new module PR [ansible-collections/community.aws/#942](https://github.com/ansible-collections/community.aws/pull/942)

**ADDITIONAL INFORMATION**
1 - I implemented Terminator in EksCluster class because it is a dependent resource and it only works with a created cluster.
2 - It was necessary to implement a waiter because if there is a fargate in process of deletion, the cluster delete fails.

### DIFF
--- a/aws/terminator/compute.py
+++ b/aws/terminator/compute.py
@@ -292,7 +292,6 @@ class EksCluster(Terminator):
             for cluster in cluster_list:
                 results.append(client.describe_cluster(name=cluster)['cluster'])
             return results
-
         return Terminator._create(credentials, EksCluster, 'eks', _build_cluster_results)
 
     @property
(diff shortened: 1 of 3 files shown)


aws/terminator/compute.py
```diff
@@ -303,7 +303,17 @@ def name(self):
...
+    def _find_eks_fargate_profile(self):
+        return self.client.list_fargate_profiles(clusterName=self.name)['fargateProfileNames']
+
     def terminate(self):
+        for profile in self._find_eks_fargate_profile():
+            self.client.delete_fargate_profile(clusterName=self.name, fargateProfileName=profile)
+            # To delete a cluster it cannot have any active fargates,
+            # so we need to delete it and wait for the end to proceed.
+            waiter = self.client.get_waiter('fargate_profile_deleted')
+            waiter.wait(clusterName=self.name, fargateProfileName=profile)
```
mattclay, line 316: Instead of waiting for the delete, why not separate the cluster deletion from the fargate profile deletion? We do that for other resources that have dependencies.

tjarra: I really don't know if its the best way to do it, but I implemented this way trying to solve 2 major problems:

1- As fargates are cluster dependent, we have no way to list/describe a fargate without informing a cluster it belongs to.

2- The process of deleting a cluster is also dependent on the existence or not of the fargate in that cluster, if you have one fargate created or in the state "deleting" (it takes about 2 minutes to delete a fargate) the deleting process fails. This also occurs between fargates, for example if you have two fargates and one is in the "deleting" state, the other can only be deleted if the first one ends, otherwise a failure occurs.
To solve this i added the waiter, this way we guarantee that you won't have any fargate created or being deleted avoiding the error.

mattclay: Any delete that takes 2 minutes is something we shouldn't be waiting on. That will block the Lambda function execution for too long. There must be some way we can check on the status of the delete on subsequent runs of the function, rather than waiting for it to complete.

tjarra: I can remove the waiter and add a try except statement on the deletions, that way if it fails we just escape and in the next lambda run it will remove it.
What do you think @mattclay?

mattclay: Yes, that should work, and is generally how we handle the situation with other resources with dependencies.

In most cases we have a class for each resource type to delete, with the shortest lifetimes defined for the ones that must be deleted first.

tjarra: I changed it to a dedicated Eks Fargate Profile class and added the try except statement. I think it's better now.

mattclay (commented):
This looks better. I'll leave it to @jillr and @gravesm for further review.

jillr (commented):
Hi @tjarram, tsting with your module PR I found a couple things that need to be added to policy/security-services.yaml to allow fargate to assume the correct roles:
```
diff --git a/aws/policy/security-services.yaml b/aws/policy/security-services.yaml
index f1928ee..93c4f25 100644
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -14,6 +14,9 @@ Statement:
       ArnLike:
         iam:PolicyArn:
           - 'arn:aws:iam::aws:policy/AWSDenyAll'
+          - 'arn:aws:iam::aws:policy/AmazonEKSServicePolicy'
+          - 'arn:aws:iam::aws:policy/AmazonEKSClusterPolicy'
+          - 'arn:aws:iam::aws:policy/AmazonEKSFargatePodExecutionRolePolicy'
           - 'arn:aws:iam::aws:policy/AWSXrayWriteOnlyAccess'
           - 'arn:aws:iam::aws:policy/IAMReadOnlyAccess'
           - 'arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore'
@@ -165,8 +168,11 @@ Statement:
     Resource:
       - 'arn:aws:iam::{{ aws_account_id }}:role/aws-service-role/autoscaling.amazonaws.com/*'
       - 'arn:aws:iam::{{ aws_account_id }}:role/aws-service-role/spot.amazonaws.com/*'
+      - 'arn:aws:iam::{{ aws_account_id }}:role/aws-service-role/eks-fargate.amazonaws.com/*'
     Condition:
       ForAnyValue:StringEquals:
         iam:AWSServiceName:
           - 'autoscaling.amazonaws.com'
           - 'spot.amazonaws.com'
+          - 'eks-fargate.amazonaws.com'
```
You can ping me on Matrix or irc (my user there is also jillr) for a re-review!

jillr (approved):
Thanks @tjarra, these will be deployed to our CI account shortly!

PR #198, opened 2022-03-12
Add terminators for DB snapshots

tremble:
The rds_instance test has been leaving snapshots behind.

I've deleted some of them so tests can pass again.


tremble:
@gravesm @jillr There's currently something in the region of 200 RDS snapshots - If you've not got the time to verify this PR right now, if possible please could you delete the snapshots.

aws/terminator/data_services.py
```diff
@@ -199,6 +199,48 @@ def terminate(self):
         self.client.delete_db_instance(DBInstanceIdentifier=self.name, SkipFinalSnapshot=True)
 
 
+class RdsDbSnapshot(Terminator):
+    @staticmethod
+    def create(credentials):
+        return Terminator._create(credentials, RdsDbSnapshot, 'rds', lambda client: client.describe_db_snapshots()['DBSnapshot'])
```
gravesm, line 205:
```suggestion
        return Terminator._create(credentials, RdsDbSnapshot, 'rds', lambda client: client.describe_db_snapshots()['DBSnapshots'])
```
I'm getting a missing key error here. Should this be `DBSnapshots`?

aws/terminator/data_services.py
```diff
@@ -199,6 +199,48 @@ def terminate(self):
...
+    def id(self):
+        return self.instance['DBSnapshotArn']
+
+    @property
+    def name(self):
+        return self.instance['DBSnapshotIdentifier']
+
+    @property
+    def created_time(self):
+        return self.instance['SnapshotCreateTime']
```
gravesm, line 217: The `SnapshotCreateTime` key does not seem to always be present. It's working for some snapshots, but I'm getting a missing key error on at least one. I wonder if we should use `DbTerminator` for this one, instead? I can't find any information on why this key would be missing.

gravesm (approved):

PR #202, opened 2022-03-24
Add policies for sns topic tagging

alinabuzachis:
SNS tagging needed by https://github.com/ansible-collections/community.aws/pull/972


marknet15 (approved):

gravesm:
@alinabuzachis This is fine in as much as it applies to SNS topics in our account, but the problem is that this is not enough to allow for `ListTagsForResource` on the third-party topic. We could add that third-party topic ARN as a resource in our policy, but because it's a cross account request, that topic would need a resource-based policy that allows us to make that request (it apparently does not).  Because the community.aws PR changes make the list tags call for every request now, this effectively breaks all the existing tests against the third-party topic.

tremble:
Tags are a bit weird, there shouldn't be a need for permissions on the other side, because you simply can't access someone else's tags, but you can attach tags that you would see to them.

tremble:
Oh, Yeah these tagging permissions are in the policy explicitly limited to the CI account.

In the `PermitReadOnlyThirdParty` stanza we'd need to add the SNS tagging permissions too.

alinabuzachis:
> Oh, Yeah these tagging permissions are in the policy explicitly limited to the CI account.
> 
> In the `PermitReadOnlyThirdParty` stanza we'd need to add the SNS tagging permissions too.

Done, thanks!

gravesm:
This is still broken for me, with the same authorization error. I also tried listing tags on the third party topic as a user with full admin privileges, and this fails in the same way. I guess I don't understand how this is supposed to work. It seems to me that all the IAM permission is doing is allowing the CI account to use SNS:ListTags on that third party resource, but that resource still would have to allow us to do so. @tremble you mention it's possible to attach tags to a third party topic. Are you doing this by just calling `tag_resource()` with that topic's ARN?

tremble:
> @tremble you mention it's possible to attach tags to a third party topic.

I've done it with other resources before.  Having had a poke about, it looks like it doesn't work for these Topics, it's probably missing a permission on the AWS side somewhere

tremble:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions

```
When you tag public or shared resources, the tags you assign are available only to your AWS account; no other AWS account will have access to those tags. For tag-based access control to shared resources, each AWS account must assign its own set of tags to control access to the resource.
```

tremble:
@alinabuzachis Looking at https://github.com/ansible-collections/community.aws/pull/972 (which I just rebased), the cross-account tagging tests have been removed, so if this PR's rebased we should be able to the this and 972 merged and closed out.

gravesm:
Closing this since I believe it is superseded by #260.

PR #205, opened 2022-04-08
Add cloud control permissions

gravesm:
These are needed to support the new amazon.cloud collection.


jillr (approved):
tested with the logs and s3 test suites - there are other errors in the tests, but the cloud control / cfn actions succeed. will deploy to prod.

PR #207, opened 2022-04-21
Policies for transit gateway

jillr:
Enable service-linked role and add terminator classes for VPC transit gateways.

Related to: https://github.com/ansible-collections/community.aws/pull/1004

### DIFF
--- a/aws/terminator/networking.py
+++ b/aws/terminator/networking.py
@@ -499,3 +499,66 @@ class NetworkFirewallRuleGroup(DbTerminator):
 
     def terminate(self):
         self.client.delete_rule_group(RuleGroupArn=self.id)
+
+
+class TransitGateway(Terminator):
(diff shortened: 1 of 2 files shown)


aws/terminator/networking.py
```diff
@@ -499,3 +499,66 @@ def name(self):
 
     def terminate(self):
         self.client.delete_rule_group(RuleGroupArn=self.id)
+
+
+class TransitGateway(Terminator):
```
tremble, line 504: There's currently something in compute.py for both of these (I put the attachment piece there because the TGW piece was there).

Probably worth deleting the ones from compute.py

jillr: ah I totally missed that - thanks!  I'll remove whatever is redudant

gravesm (approved):

PR #208, opened 2022-04-28
add rds policy for CopyDBSnapshot

jatorcasso:
needed by https://github.com/ansible-collections/community.aws/pull/1078

### DIFF
--- a/aws/policy/data-services.yaml
+++ b/aws/policy/data-services.yaml
@@ -43,12 +43,7 @@ Statement:
       - elasticache:DeleteCacheCluster
       - elasticache:DeleteCacheSecurityGroup
       - elasticache:DeleteCacheSubnetGroup
-      - elasticache:DescribeCacheClusters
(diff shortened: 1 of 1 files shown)


jatorcasso:
recheck

gravesm:
Unless I'm mistaken, I don't think we can support copying a db snapshot to another region, because the terminator is only going to clean up snapshots in us-east-1. @jillr can you confirm?

jatorcasso:
> Unless I'm mistaken, I don't think we can support copying a db snapshot to another region, because the terminator is only going to clean up snapshots in us-east-1. @jillr can you confirm?

ah, okay. that makes sense! If that's the case, I can comment out those integration tests and leave a comment.

jatorcasso:
@gravesm updated integration tests to copy snapshots in the same region 👍🏻

gravesm (approved):

PR #209, opened 2022-04-29
Allow logs:ListLogDeliveries

tremble:
Working on https://github.com/ansible-collections/community.aws/pull/1107 I've discovered that even if you don't configure logging, you need to be able to **list** the logging configuration to delete NetworkFirewall firewalls


tremble:
@gravesm - If you get chance, please could you look at this one.  I've got some tests I want to upload, but doing so without this change will result in a build up of Firewall resources, subnets and VPCs in the CI account.

gravesm:
Sorry for the delay. Changes should go out shortly.

PR #210, opened 2022-05-13
Policies for ecs_service, ecs_cluster, ecs_taskdefinition, ecs_task

alinabuzachis:
Policies for ecs_service, ecs_cluster, ecs_taskdefinition, ecs_task

Needed for https://github.com/ansible-collections/community.aws/pull/1145


aws/terminator/compute.py
```diff
@@ -618,3 +618,82 @@ def created_time(self):
...
+                return []
+
+            return self.client.describe_services(cluster=self.name, services=names)['services']
+
+        container_instances = _paginate_container_instance_results()
+        for name in container_instances:
+            # If there are tasks running on a container instance, stop them first
+            tasks = _paginate_task_results(name['containerInstanceArn'])
+            for task in tasks:
+                self.client.stop_task(cluster=self.name, task=task['taskArn'])
```
markuman, line 690:
```suggestion
            tasks = _paginate_task_results(name['containerInstanceArn'])
            for task in tasks:
                self.client.stop_task(cluster=self.name, task=task['taskArn'])
```

Unless the desired count of the belonging service is not set  to 0, the service might spawn new tasks to replace the stopped ones.

markuman: maybe flipping solves this issue in a lazy way?

1. delete the service
2. delete all left running tasks (_like `killall -9`_)
3. delete the cluster

gravesm: Do the tasks need to be individually deleted at all? Won't deregistering the container instance also remove the tasks?

markuman: If the ecs_tasks runs on ec2, yes. deleteing the instance, kills also the ecs tasks.  
but if it's serverless (fargate), there is no otherway to stop/kill the tasks.  

but basically you did it that way, it you won't wait for the service to become inactive.

imo we've got two choices.  
a) delete service and wait until it's gone. forward delete cluster  
b) delete service, kill pending tasks, forward delete cluster

aws/terminator/compute.py
```diff
@@ -618,3 +618,82 @@ def created_time(self):
...
+
+            self.client.deregister_container_instance(containerInstance=name['containerInstanceArn'])
+
+        # If there are running services, delete them first
+        services = _paginate_service_results()
+        for name in services:
+            self.client.delete_service(cluster=self.name, service=name['serviceName'], force=True)
+
+        self.client.delete_cluster(cluster=self.name)
```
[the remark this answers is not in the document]
gravesm: We would typically avoid using waiters on the lambda to prevent timeouts, especially for anything that might take more than a minute or two. The ideal way to handle a scenario like this would be to create terminators for the cluster and any dependent resources, and increase the age limit on the cluster terminator. This is complicated by the fact that doing anything with services or container instances also requires the cluster name.

aws/terminator/compute.py
gravesm: I feel like the first option seems like less work.

markuman: https://github.com/mattclay/aws-terminator/pull/210#discussion_r879421673

Then we must implement waiters for the service.

gravesm: If a service truly cannot be deleted until all tasks are stopped, I don't think we can use a waiter. A task could take up to 2 minutes to shutdown, plus however long it takes for amazon to issue the shutdown request. We can't block the lambda for that long.

mattclay: The terminator must not wait on operations to complete, as that will block it from performing other actions. Resource age limits should be used to control the order of deletions when dependencies are involved.

markuman: @gravesm @mattclay another possibility is to wrap `delete_cluster` into try/except.  
imo, the ecs services and tasks will be stopped/deleted. maybe it takes some time until they gone an the `delete_cluster` will fail because of that. but it doesn't matter. because the next run of the terminator will delete the empty cluster then. and the ecs cluster itself has no fees

jillr: We wouldn't want to ignore the exception forever though.  So perhaps, one class with the default `age_limit` that catches and ignores `ClusterContainsServicesException` and `ClusterContainsTasksException` on `delete_cluster()`.  Then a second class, which has an `age_limit` 10 minutes longer than the default, that only tries to delete clusters (and that one doesn't ignore those exceptions, so that errors still surface to us in the monitoring).

I'm assuming 10 minutes should be more than long enough, even on a bad day, for all tasks and services to be 100% gone.

markuman: That sounds good!

alinabuzachis: @markuman I tried to create the second class, @markuman can you check please and let me know if this is what you meant? Thanks.

markuman: LGTM.
What's about the failing CI. I cannot find any usefull error message.

alinabuzachis: @markuman it needs to be refactored in some way because the file's dimension exceeds the limit. Will try to see what I can do for this thing.

gravesm:
Since we've exceeded the policy size on compute, I would suggest moving the ECS policies to the paas policy.

aws/terminator/paas.py
```diff
@@ -20,3 +22,121 @@ def ignore(self) -> bool:
...
+
+        container_instances = _paginate_container_instance_results()
+        for name in container_instances:
+            # If there are tasks running on a container instance, stop them first
+            tasks = _paginate_task_results(name['containerInstanceArn'])
+            for task in tasks:
+                self.client.stop_task(cluster=self.name, task=task['taskArn'])
+
+        # If there are running services, delete them first
+        services = _paginate_service_results()
```
gravesm, line 101: Services should be the first things that are deleted, otherwise, any deleted tasks may just be restarted again.

aws/terminator/paas.py
```diff
@@ -20,3 +22,121 @@ def ignore(self) -> bool:
...
+        for name in services:
+            self.client.delete_service(cluster=self.name, service=name['serviceName'], force=True)
+
+        # Deregister container instances
+        for name in container_instances:
+            self.client.deregister_container_instance(containerInstance=name['containerInstanceArn'])
+
```
gravesm, line 108: Before deleting the cluster, you will also need to delete all the tasks. In the case of fargate, there are no container instances and you may have tasks running that aren't managed by a service. I'm also not clear on whether using `force=True` when deleting a service also deletes the tasks or just orphans them. Either way, there should be a step that stops all tasks.

gravesm (commented):
A couple small fixes, but otherwise this LGTM.

aws/terminator/paas.py
```diff
@@ -20,3 +22,131 @@ def ignore(self) -> bool:
...
+            names = client.get_paginator('list_clusters').paginate(
+                PaginationConfig={
+                    'PageSize': 100,
+                }
+            ).build_full_result()['clusterArns']
+
+            if not names:
+                return []
+
+            return client.describe_clusters(name=names)['clusters']
```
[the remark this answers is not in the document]
alinabuzachis: Done! Thanks.

aws/terminator/paas.py
```diff
@@ -20,3 +22,131 @@ def ignore(self) -> bool:
...
```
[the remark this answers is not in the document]
alinabuzachis: Done! Thanks.

tremble:
@alinabuzachis @gravesm Where do we stand on getting this merged?

gravesm:
Hey, I'll try to get to this early next week. There are a lot of parts covered by this PR and I'd like to set aside some time to test it.

markuman:
Just a note, the `ecs_cluster` integration test is currently broken.

gravesm:
@tremble I'm running into a lot of problems with the ecs_cluster tests. There may be other things that are broken, as @markuman referred to, but the biggest problem I have so far is that the tests seem to be intended to provision containers on EC2 instances, but then the service definitions are using Fargate provisioning.

What I need in order to get this merged is a working test suite. I'm willing to work on getting the permissions correct, but I need a PR I can test against that is otherwise working.

markuman:
> What I need in order to get this merged is a working test suite. I'm willing to work on getting the permissions correct, but I need a PR I can test against that is otherwise working.

I can try to target it this week.

PR #212, opened 2022-06-01
Add policy for rds_cluster_snapshot copy_rds_cluster_snapshot

alinabuzachis:
Add policy for rds_cluster_snapshot copy_rds_cluster_snapshot


gravesm:
Your community.aws PR is failing for me before it even gets to the copy step. The CI seems to be failing at the same point. Would it be possible to address those issues, first?

alinabuzachis:
> Your community.aws PR is failing for me before it even gets to the copy step. The CI seems to be failing at the same point. Would it be possible to address those issues, first?

It should be solved now. Thanks @gravesm

PR #217, opened 2022-06-16
Allow GetParameter on SSM resources we've created

tremble:
https://github.com/ansible-collections/amazon.aws/pull/873 - some of the later tests will probably still fail due to the same issue as https://github.com/ansible-collections/amazon.aws/pull/810


tremble:
@gravesm Any chance of another quick and easy review - we need both getParameter **and** getParameters

gravesm (approved):

PR #220, opened 2022-07-22
Allow policy validation API access

tremble:
AWS now have an API for validating and linting policy documents:

https://docs.aws.amazon.com/access-analyzer/latest/APIReference/API_ValidatePolicy.html

https://github.com/ansible-collections/community.aws/pull/1370

### DIFF
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -66,6 +66,7 @@ Statement:
   - Sid: AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees
     Effect: Allow
     Action:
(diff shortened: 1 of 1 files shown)


tremble:
@gravesm If you've got the cycles - another minor policy update for your review

gravesm (approved):
Neat. I wonder if this would be useful at all for this repo?

tremble:
> Neat. I wonder if this would be useful at all for this repo?

I was thinking the same thing :)  Only complication is that it needs credentials to be able to run, which would (if possible) want to be exposed to the CI

PR #222, opened 2022-08-22
Allow create db cluster parameter group for rds_cluster modules

imesias:
To Allow expanded integration tests on.

https://github.com/ansible-collections/community.aws/pull/1417

Expand the IAM policy to allow create/delete rds cluster parameter groups.

### DIFF
--- a/aws/terminator/data_services.py
+++ b/aws/terminator/data_services.py
@@ -196,7 +196,7 @@ class RdsDbClusterParameterGroup(DbTerminator):
         return self.name.startswith('default')
 
     def terminate(self):
-        self.client.delete_db_parameter_group(DBClusterParameterGroupName=self.name)
+        self.client.delete_db_cluster_parameter_group(DBClusterParameterGroupName=self.name)
 
 
 class RdsDbInstance(DbTerminator):
--- a/aws/terminator/data_services.py
+++ b/aws/terminator/data_services.py
@@ -181,7 +181,8 @@ class RdsDbParameterGroup(DbTerminator):
 class RdsDbClusterParameterGroup(DbTerminator):
     @staticmethod
     def create(credentials):
-        return Terminator._create(credentials, RdsDbClusterParameterGroup, 'rds', lambda client: client.describe_db_cluster_parameter_groups()['DBClusterParameterGroups'])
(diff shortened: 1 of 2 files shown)


imesias:
image

@markuman @alinabuzachis Receiving an odd error from CI, apparently there is an Ubuntu brownout and Azure can't process the request.

aws/terminator/data_services.py
```diff
@@ -178,6 +178,27 @@ def terminate(self):
...
+    def create(credentials):
+        return Terminator._create(credentials, RdsDbClusterParameterGroup, 'rds', lambda client: client.describe_db_cluster_parameter_groups()['DBClusterParameterGroups'])
+
+    @property
+    def id(self):
+        return self.instance['DBClusterParameterGroupArn']
+
+    @property
+    def name(self):
+        return self.instance['DBClusterParameterGroup']
```
gravesm, line 192:
```suggestion
        return self.instance['DBClusterParameterGroupName']
```

gravesm (changes requested):

aws/terminator/data_services.py
imesias: @gravesm Ah, thanks for spotting this. hmm, If I reformat line 184 it kinda breaks the consistency in the module. Should I add reformat or is there a way to tell pycodestyle to ignore that particular line by use of a code? Or should I bump the line length in this case.

imesias: Oh actually withdraw previous statements will re-format I see line 227  is a reformatted line lower down :-) Will reformat and submit shortly. Thanks for your assistance. Much appreciated.

imesias:
Tests are now passing however, now encountering the following assuming due to the addition of the new policy statements.
image

gravesm:
This looks fine, but I'm going to have to readjust some things to get the policy length test to pass. Most of our policies are so large now that they bump up against the size limit imposed by AWS. I'll just have to try and consolidate some things.

jillr (approved):
I haven't tested the new class, but the policy refactor LGTM. If you're +1 the new class, @gravesm, then I am too.

gravesm (approved):

PR #225, opened 2022-08-26
Add missing cluster parameter group resource type

imesias:
Add missing cluster parameter group resource type


imesias:
@gravesm Apologies for the delay

gravesm:
Tests are now passing locally for me. Will get these changes deployed shortly.

PR #228, opened 2022-09-21
Added policies for EFS backup

akamanzi:
Added required policies for `community.aws` collection PR https://github.com/ansible-collections/community.aws/pull/1502


gravesm (commented):
I believe this is potentially going to result in restore points being created. We will need an associated terminator class to delete restore points and backup vaults.

PR #230, opened 2022-09-27
update terminator.yml to use boto3 modules (iam_role)

abikouo:
``aws/terminator.yml`` is using the ``community.aws.iam`` module which was previously deprecated, and then removed in 3.0.0.
update playbook to use ``community.aws.iam_role`` module

### DIFF
--- a/aws/terminator.yml
+++ b/aws/terminator.yml
@@ -28,17 +28,11 @@
         that: "aws_account_id == lambda_account_id"
     - name: create iam role for terminator functions
       tags: iam
-      iam:
-        iam_type: role
+      community.aws.iam_role:
         name: "{{ api_name }}-terminator-{{ stage }}"
+        description: "iam role for terminator functions"
         state: present
-        trust_policy:
-          Version: '2012-10-17'
-          Statement:
-            - Action: sts:AssumeRole
-              Effect: Allow
-              Principal:
-                Service: lambda.amazonaws.com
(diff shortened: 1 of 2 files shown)


aws/terminator.yml
```diff
@@ -28,17 +28,11 @@
...
+        description: "iam role for terminator functions"
         state: present
-        trust_policy:
-          Version: '2012-10-17'
-          Statement:
-            - Action: sts:AssumeRole
-              Effect: Allow
-              Principal:
-                Service: lambda.amazonaws.com
+        assume_role_policy_document: '{"Version": "2012-10-17", "Statement": {"Action": "sts:AssumeRole", "Principal": {"Service": "lambda.amazonaws.com"}, "Effect":"Allow"}}'
```
tremble, line 35: I'd recommend passing this as a dict rather than a JSON string, it's easier to work with.

tremble (approved):

aws/terminator.yml
```diff
@@ -28,17 +28,17 @@
...
+        assume_role_policy_document:
+          Version: "2012-10-17"
           Statement:
-            - Action: sts:AssumeRole
-              Effect: Allow
-              Principal:
-                Service: lambda.amazonaws.com
+            Action: "sts:AssumeRole"
+            Principal:
+              Service": "lambda.amazonaws.com"
```
gravesm, line 40:
```suggestion
              Service: "lambda.amazonaws.com"
```

gravesm (approved):

PR #232, opened 2022-09-28
Use unrestricted permissions for listing certs

gravesm:
There's no way to filter certs by Arn when making a list call, so the permissions need to allow for unrestricted listing.


tremble (approved):
List is reasonably safe.

PR #233, opened 2022-10-03
Remove dependency with `mattclay.aws` collection

abikouo:
`aws/terminator.yml` and `aws/deploy-test-policy.yml` playbooks use `mattclay.aws` collection, we want to remove this and use directly supported collections `community.aws` and `amazon.aws`


abikouo:
> I am not sure that we need to prioritize removing the dependency on mattclay.aws. If there are modules in the supported amazon.aws collection that provide the exact same functionality as modules in mattclay.aws, we can look at replacing those. We shouldn't be vendoring copies of modules from mattclay.aws just to remove that dependency, though.

This is exactly what has been done. The mattclay.aws modules have been replaced by those from supported collection with the same feature. Nothing has been copied, `terminator_lambda_package.py` was copied by can be replaced by the `community.general.archive` module doing the same

gravesm:
> Nothing has been copied, terminator_lambda_package.py was copied

This is my problem. Why copy anything at all? There is no pressing need to remove the dependency on mattclay.aws.

abikouo:
> > Nothing has been copied, terminator_lambda_package.py was copied
> 
> This is my problem. Why copy anything at all? There is no pressing need to remove the dependency on mattclay.aws.

I will turn `terminator_lambda_package` into `community.general.archive`

aws/terminator-policy.json
```diff
@@ -17,7 +17,7 @@
         {
             "Effect": "Allow",
             "Action": "sts:AssumeRole",
-            "Resource": "arn:aws:iam::{{ test_account_id }}:role/{{ api_name }}-test-{{ stage }}"
+            "Resource": "arn:aws:iam::{{ test_account_id }}:role/{{ iam_role_name }}"
```
gravesm, line 20: This is not the same and needs to remain `{{ api_name }}-test-{{ stage }}`

aws/terminator.yml
```diff
@@ -39,100 +50,100 @@
...
+        name: "{{ lambda_layer_name }}"
+        description: "Python requirements for {{ lambda_function_name }}"
         compatible_runtimes:
-          - python3.9
-        path: "{{ packaging_dir }}/terminator-requirements.zip"
+          - "{{ lambda_python_version }}"
+        content:
+          zip_file: "{{ packaging_dir }}/terminator-requirements.zip"
         license_info: GPL-3.0-only
-        region: "{{ aws_region }}"
```
gravesm, line 85: Don't we still need the region? Can we add this to `module_defaults`?

requirements.yml
```diff
@@ -5,3 +5,6 @@ collections:
     version: '3.3.0'
   - name: community.general
     version: '>=6.0.0'
+  - name: amazon.aws
+    version: '>5.0.0'
+  - name: community.general
```
gravesm, line 10: This last line should be removed. Sorry, left over from my rebase.

aws/vars/main.yml
```diff
@@ -0,0 +1,20 @@
+aws_region: "us-east-1"
+stage: "prod"
+terminator_policy: "{{ lookup('template', 'terminator-policy.json') }}"
+packaging_dir: "{{ playbook_dir }}/../.cache/packaging"
+virtualenv_requirement_file: "{{ playbook_dir }}/requirements.txt"
+lambda_python_version: "python3.7"
```
gravesm, line 6: This should be `python3.9` now.

gravesm (commented):
I went ahead and rebased this. Sorry for the delay. There are a few small things that need addressing, but otherwise it looks good.

aws/terminator.yml
```diff
@@ -39,100 +50,100 @@
...
         virtualenv: "{{ packaging_dir }}/terminator-requirements/python"
-        virtualenv_python: python3.9
+        virtualenv_python: "{{ lambda_python_version }}"
+      check_mode: false
+
     - name: package terminator requirements
       tags: lambda
-      lambda_package:
-        src: "{{ packaging_dir }}/terminator-requirements"
+      archive:
```
mattclay, line 73: This isn't a full replacement for the `lambda_package` module. The generated archive changes based on the timestamps of the included files, which results in a new Lambda version being created even when the file contents are unchanged.

mattclay (changes requested):

aws/terminator.yml
abikouo: updated

aws/terminator.yml
abikouo: Fixed.
Idempotency check for the lambda layer archive is assumed by downloading the one stored in AWS and compare it with the local one in a deterministic way (setting a constant time value)

mattclay: Why add a custom module to do the comparison instead of using the existing module to create the archive?

abikouo: existing I did not found a way in the module to create archive in a deterministic way, so we will always have a diff between the sha from aws and the local one, please suggest if there is a way to solve that

mattclay: Sorry, I wasn't specific enough. I was asking why replace `mattclay.aws.lambda_package`?

abikouo: `mattclay.aws.lambda_package` works with `mattclay.aws.lambda_layer` because the lambda_layer from that collection expects an archive with a deterministic timestamp to ensure idempotency, this behavior is not part of the `amazon.aws` collection. However I can keep `amazon.aws.lambda_package` and perform the logic on the playbook, this is almost what I have done but without the dependency of the `mattclay.aws` collection

mattclay: What's the reason for wanting to remove the dependency on the `mattclay.aws` collection to begin with?

gravesm: > I think the intend was to migrate to a collection supported by the cloud team. @jillr could you please clarify

This was never the intent. There was a need to update `community.aws`, in particular migrating to the newer `iam_role` module. There's never been a problem with the `mattclay.aws` collection. If there are modules from the supported `amazon.aws` collection that can fully replace modules from the `mattclay.aws` collection, we can consider those, but that is not apparently the case here with the `lambda_package`.

abikouo: ok then I will keep lambda_package from `mattclay.aws`

mattclay: FYI - I've updated the `mattclay.aws` collection to version 3.0.0. It should now play nice with `amazon.aws` 5.x.

gravesm:
I would like to either close this PR or get it to a state that it can be merged. With `mattclay.aws` upgraded to now work with newer versions of `amazon.aws` and `community.aws`, I'm not sure the changes here are necessary. There was a need to move off of older, unsupported modules, but that's no longer a problem.

PR #234, opened 2022-10-03
add support for lambda layer

abikouo:
add the following permission for lambda:
- ListLayers
- ListLayerVersion
- PublishLayerVersion
- DeleteLayerVersion

Required for https://github.com/ansible-collections/amazon.aws/pull/1095

### DIFF
--- a/aws/terminator/paas.py
+++ b/aws/terminator/paas.py
@@ -20,3 +20,21 @@ class LambdaEventSourceMapping(DbTerminator):
 
     def terminate(self):
         self.client.delete_event_source_mapping(UUID=self.id)
+
+
+class LambdaLayers(DbTerminator):
+    @staticmethod
+    def create(credentials):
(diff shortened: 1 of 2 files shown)


gravesm:
This will need an associated `LambdaLayerVersion` terminator class.

aws/terminator/paas.py
```diff
@@ -20,3 +20,21 @@ def ignore(self) -> bool:
 
     def terminate(self):
         self.client.delete_event_source_mapping(UUID=self.id)
+
+
+class LambdaLayers(DbTerminator):
```
gravesm, line 25: Instead of making this a `DbTerminator` I would just make it a `Terminator` and use `self.instance['LatestMatchingVersion']['CreatedDate']` as the `created_time`.

abikouo: done

gravesm (approved):

PR #238, opened 2022-11-01
use python 3.9 by default

goneri:
- upgrade to Python 3.8.
- Use ubuntu-22.04 for Azure Pipeline.


goneri:
See: https://github.com/mattclay/aws-terminator/pull/239

PR #239, opened 2022-11-02
refresh the sanity tests and use tox

goneri:
- Use `tox` to run the sanity tests
- Upgrade the default versions
- Break the dependency on Python3.7
- Adjust some issues in the code to be compliant with the update tests

These dependencies should be merged first:
- https://github.com/mattclay/aws-terminator/pull/246
- https://github.com/mattclay/aws-terminator/pull/247


README.md
```diff
@@ -1,3 +1,17 @@
 # AWS Terminator
 
 An AWS Lambda function for cleaning up AWS resources.
+
+## Run the sanity tests
+
+We use `tox` to run the sanity tests:
+
+``` sh
```
webknjaz, line 9: If there's a prompt, the console lexer should be used for proper highlighting. Otherwise, drop the prompt.
```suggestion
```console
```

aws/cleanup.py
```diff
@@ -50,7 +50,7 @@ def default_ctor(_dummy, tag_suffix, node):
 
     yaml.add_multi_constructor('', default_ctor)
 
-    with open(config_path) as config_fd:
+    with open(config_path, encoding="UTF-8") as config_fd:
```
webknjaz, line 53: I rarely ever see the encodings uppercased in Python. I suppose, let's stick to the convention:
```suggestion
    with open(config_path, encoding='utf-8') as config_fd:
```

tox.ini
```diff
@@ -0,0 +1,32 @@
+[tox]
+skipsdist=True
+envlist=pycodestyle,pylint,yamllint
+
+[testenv]
+deps =
+  -cconstraints.txt
+  -raws/requirements.txt
+
+[testenv:pylint]
```
webknjaz, line 10: It's typically nicer to consolidate various linting tools under pre-commit, like this: https://github.com/ansible/pylibssh/blob/bfc1e97/.pre-commit-config.yaml + https://github.com/ansible/pylibssh/blob/bfc1e97/tox.ini#L296-L309.

tox.ini
```diff
@@ -0,0 +1,32 @@
...
+  pycodestyle
+commands = pycodestyle --config pycodestyle.ini aws
+
+[testenv:yamllint]
+deps =
+  {[testenv]deps}
+  yamllint
+commands = yamllint --config-file .yamllint aws hacking
+
+[testenv:policy]
```
webknjaz, line 28: It's a good idea to provide a `description` definition so that it's available in the `tox -av` output.

tox.ini
```diff
@@ -0,0 +1,32 @@
...
+deps =
+  {[testenv]deps}
+  ansible-core
+commands = ansible-playbook -i localhost hacking/aws_config/test-policies.yml
```
webknjaz, line 32: I'd use full paths where possible:
```suggestion
commands = ansible-playbook -i localhost {toxinidir}/hacking/aws_config/test-policies.yml
```

tox.ini
goneri: Thanks for the suggestion. It's a good idea but slightly out of scope for this PR.

goneri:
Thank you @webknjaz for the review!

tox.ini
```diff
@@ -0,0 +1,36 @@
+[tox]
+skipsdist=True
+envlist=pycodestyle,pylint,yamllint
+
+[test-deps]
+deps =
+  -cconstraints.txt
+  -raws/requirements.txt
+
+[test-deps:pylint]
```
webknjaz, line 10: This should remain 
```suggestion
[testenv:pylint]
```
For tox to notice it.

abikouo (approved):

Makefile
```diff
@@ -9,12 +9,11 @@ default:
 .PHONY: test-all
 test-all: test
 	make test -C aws
-	make test -C hacking
+	tox
```
mattclay, line 12: Between the switch to `tox` and previous changes that were made, it appears several files are no longer needed:

* this `Makefile`
* all 3 `test-requirements.txt` files
* the empty `hacking/requirements.txt` file

aws/terminator/__init__.py
```diff
@@ -270,7 +270,7 @@ def is_vpc_default(self, vpc_id: str) -> bool:
 class DbTerminator(Terminator):
     """Base class for classes which find and terminate AWS resources with age tracked via DynamoDB."""
     def __init__(self, client: botocore.client.BaseClient, instance: typing.Dict[str, typing.Any]):
-        super(DbTerminator, self).__init__(client, instance)
+        super().__init__(client, instance)
```
mattclay, line 273: It would be nice to review and discuss the change to `tox` independently of the code fixes in `aws/terminator/`, which shouldn't need much discussion. Could you put this, and the other fixes, into a separate PR?

aws/requirements.txt
```diff
@@ -1,12 +1,12 @@
...
+jmespath==1.0.1
 python-dateutil==2.8.2
-requests==2.27.1
-s3transfer==0.5.2
+requests==2.28.1
+s3transfer==0.6.0
 six==1.16.0
-urllib3==1.26.8
+urllib3==1.26.12
+PyYAML==6.0
```
mattclay, line 12: The `PyYAML` requirement is only used for the `cleanup.py` script, so shouldn't be included here. Including it here will put it into the venv which gets packaged up for the Lambda function, which isn't needed.

Perhaps the requirement should go into a separate dev-requirements.txt file.

aws/Makefile
```diff
@@ -31,22 +31,3 @@ terminator_lambda:
 .PHONY: test_policy
 test_policy:
 	$(RUN_TEST_POLICY_PLAYBOOK)
-
```
mattclay, line 34: It looks like `VAULT_ARGS` above is unused. Also, the "Running Tests" section in the help above is obsolete.

Alternatively, what do you think of converting the rest of this `Makefile` over to running through `tox`?

mattclay (changes requested):

aws/terminator/__init__.py
goneri: I've just opened this PR with the Python changes. It has to be merged before this PR or the tests will fail: https://github.com/mattclay/aws-terminator/pull/246

aws/Makefile
goneri: I've done the clean up. It would be nice indeed to move everything to `tox`. But I would prefer to do it later in another PR later once the Ubuntu22.04/Python3.9 upgrade is done.

aws/requirements.txt
```diff
@@ -1,12 +1,17 @@
-boto3==1.21.10
-botocore==1.24.10
-certifi==2021.10.8
-charset-normalizer==2.0.12
-docutils==0.18.1
-idna==3.3
-jmespath==0.10.0
+# note: The dependencies will be include in the Lamda package.
```
webknjaz, line 1:
```suggestion
# note: The dependencies will be included in the Lambda package.
```

mattclay (approved):

jillr (approved):

PR #240, opened 2022-11-02
DNM - Python3

goneri:
Depends-On: https://github.com/mattclay/aws-terminator/pull/239


aws/terminator.yml
```diff
@@ -52,34 +52,34 @@
       pip:
         requirements: "{{ playbook_dir }}/requirements.txt"
         virtualenv: "{{ packaging_dir }}/terminator-requirements/python"
-        virtualenv_python: python3.7
+        virtualenv_python: python3.9
```
abikouo, line 55: we should rather define this in a dedicated to simplify code updates, I have done it in another PR

abikouo (approved):

aws/terminator.yml
goneri: Oh sorry for the confusion, this PR is not ready for review. It's a series of commits that I want to submit in independent PR.

PR #245, opened 2022-11-15
Terminator policies for CloudFront modules

abikouo:
Needed for PR https://github.com/ansible-collections/community.aws/pull/1596


gravesm:
Could you add the link in the description to the PR in the amazon collection that this change is for?

abikouo:
> Could you add the link in the description to the PR in the amazon collection that this change is for?

PR has been updated accordingly

aws/policy/paas.yaml
```diff
@@ -24,6 +24,14 @@ Statement:
   - Sid: AllowResourceRestrictedActionsWhichIncurNoFees
     Effect: Allow
     Action:
+      - cloudfront:CreateDistribution
+      - cloudfront:CreateDistributionWithTags
+      - cloudfront:DeleteDistribution
+      - cloudfront:UpdateDistribution
+      - cloudfront:TagResource
+      - cloudfront:UntagResource
+      - cloudfront:ListTagsForResource
```
jillr, line 33: Anything with a list, get, or describe should be specified as a wildcard.  This helps us stay under the maximum character length per policy that AWS imposes.
For example,
```suggestion
      - cloudfront:List*
```
The wildcard line probably needs to be in the AllowUnrestrictedResourceActionsWhichIncurNoFees block since some of the list actions are not resource restricted.

aws/policy/paas.yaml
```diff
@@ -24,6 +24,14 @@ Statement:
   - Sid: AllowResourceRestrictedActionsWhichIncurNoFees
     Effect: Allow
     Action:
+      - cloudfront:CreateDistribution
+      - cloudfront:CreateDistributionWithTags
```
jillr, line 28: I can't find this action on https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazoncloudfront.html
Should it be `CreateStreamingDistributionWithTags`?  We'll need a `'arn:aws:cloudfront::{{ aws_account_id }}:streaming-distribution/*'` resource for the streaming distro policies.

abikouo: I added the permissions for the streaming distribution.
I was trying to add permission for this https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_CreateDistributionWithTags.html, it is strange that it is not part of the documentation

aws/policy/paas.yaml
```diff
@@ -24,6 +24,14 @@ Statement:
   - Sid: AllowResourceRestrictedActionsWhichIncurNoFees
     Effect: Allow
     Action:
+      - cloudfront:CreateDistribution
+      - cloudfront:CreateStreamingDistribution
```
jillr, line 28: The create/update actions should always be under a Sid that has "ActionsWhichIncurFees" in the name.  Right now everything is in an "ActionsWhichIncurNoFees" Sid and only separated by whether they're resource restricted, the relevant actions should be moved.

For whatever reason, `create_distribution` only works when I remove the resource restriction (despite what the docs suggest).  IMO this is fine, so let's remove the resource restriction on the create actions (by moving these actions to an appropriate Sid).

aws/terminator/application_services.py
```diff
@@ -367,3 +367,41 @@ def name(self):
 
     def terminate(self):
         self.client.delete_document(Name=self.name)
+
+
+class CloudFrontDistribution(Terminator):
```
jillr, line 372: When there are no distributions to be listed the API returns an empty dictionary instead of an empty `StreamingDistributionList` or empty list of `Items`.  Meaning,
 
`client.get_paginator('list_streaming_distributions').paginate().build_full_result()['StreamingDistributionList']['Items']` returns a KeyError if there are no distributions (same for streaming distributions).  

For example, 
```
response = client.get_paginator('list_streaming_distributions').paginate().build_full_result() 
print(response)
{}
```
We'll need to execute the paginator, and then check to see if it contains any keys before deciding what to return in the create method of each terminator class.

aws/terminator/application_services.py
```diff
@@ -367,3 +367,41 @@ def name(self):
...
+    @property
+    def created_time(self):
+        return self.instance['LastModifiedTime']
+
+    @property
+    def name(self):
+        return self.instance['DomainName']
+
+    def terminate(self):
+        self.client.delete_distribution(Id=self.instance['Id'])
```
jillr, line 388: Distributions need to be disabled before they can be deleted.  The update event (setting Enabled=False) must then fully propagate across the CDN before the distro can be deleted.  https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/HowToDeleteDistribution.html

This logic will need to be built into the terminate method for each class.  ie, check if the distro is Enabled=True, disable it if it is and move on.  If the distro is already Enabled=False, check if the Status=Deployed.  When both Enabled=False and State=Deployed, try to delete.

jillr (changes requested):
Apologies I missed some of this on my previous look.
The policies are in paas.yml but the terminator class is in application-services.py - these should be in the same grouping in both places to avoid confusion.

There may be a bug in the plugin code, the integration tests fail at:
```
TASK [cloudfront_distribution : Update the distribution with tags] **************************************************************************************************************************************************************************
task path: /root/ansible_collections/community/aws/tests/output/.tmp/integration/cloudfront_distribution-5u0dnd61-ÅÑŚÌβŁÈ/tests/integration/targets/cloudfront_distribution/tasks/main.yml:34
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: get_distribution() got an unexpected keyword argument 'id'
fatal: [testhost]: FAILED! => {"boto3_version": "1.18.0", "botocore_version": "1.21.0", "changed": false, "msg": "Error validating distribution_id from alias, aliases and caller reference: get_distribution() got an unexpected keyword argument 'id'", "resource_actions": []}
```

I have the policies deployed to the dev area of the CI account (with all of the policies moved to a ResourceUnrestricted Sid for the sake of testing).  You can test your PR locally against the CI account like `ansible-test integration -vv cloudfront_distribution --remote-stage dev  --docker` with the deployed dev policy.  We can figure out the resource restrictions in another iteration.

aws/policy/paas.yaml
abikouo: updated

aws/terminator/application_services.py
abikouo: done!! Thanks

aws/policy/paas.yaml
```diff
@@ -24,6 +29,11 @@ Statement:
   - Sid: AllowResourceRestrictedActionsWhichIncurNoFees
     Effect: Allow
     Action:
+      - cloudfront:DeleteDistribution
+      - cloudfront:UpdateDistribution
+      - cloudfront:TagResource
+      - cloudfront:UntagResource
+      - cloudfront:DeleteStreamingDistribution
```
gravesm, line 36: Small nit, but could you make sure that all the added rules are ordered alphabetically within the Sid? This hasn't been done consistently, but it can get really difficult to maintain if we don't do this.

aws/terminator/paas.py
```diff
@@ -44,3 +44,86 @@ def created_time(self):
...
+
+        distribution = self.client.get_distribution(Id=self.Id)
+        ETag = distribution['ETag']
+        distribution = distribution['Distribution']
+        if distribution.get('Status') == "Deployed":
+            if distribution['DistributionConfig']['Enabled']:
+                distribution['DistributionConfig']['Enabled'] = False
+                self.client.update_distribution(DistributionConfig=distribution['DistributionConfig'], Id=self.Id, IfMatch=ETag)
+                # wait until the distribution is deployed
+                wait_until_deployed(self.client, self.Id, 'distribution_deployed')
```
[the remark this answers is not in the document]
abikouo: That is something I was suspecting. I will update to just disabled when it is not.
Thanks

aws/policy/paas.yaml
abikouo: done

gravesm (approved):

PR #249, opened 2022-11-28
use Python 3.9

goneri:
Upgrade the default Python version. Run the `policy` check by default now we've got a Python version that is compatible with an up to date `ansible-core` release.

### DIFF
--- a/aws/Makefile
+++ b/aws/Makefile
@@ -1,4 +1,4 @@
-PYTHON3 ?= python3.7
+PYTHON3 ?= python3.9
 
 EXTRA_VARS=ansible_connection=local ansible_python_interpreter=$(shell which python)
 RUN_TEST_POLICY_PLAYBOOK=ansible-playbook -i localhost, -e '$(EXTRA_VARS)' deploy-test-policy.yml $(FLAGS)
--- a/aws/README.md
+++ b/aws/README.md
@@ -70,7 +70,7 @@ Warning: Always use the --check (or -c) flag and the --target flag to avoid acci
 It is safest to use `cleanup.py` in an empty/dev account.
 
 To start using `cleanup.py` you will need to:
-* Use Python 3.7+
+* Use Python 3.9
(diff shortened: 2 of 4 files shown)


goneri:
Needs https://github.com/mattclay/aws-terminator/pull/248 first.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 skipsdist=True
-envlist=pycodestyle,pylint,yamllint
+envlist=pycodestyle,pylint,yamllint,policy
```
mattclay, line 3: Is this change related to the Python 3.9 change?

goneri: Yes, we were not able to run the policy check anymore with Ubuntu18.04/Python3.7. I will mention this change in the PR description.

goneri: Ok, this raise a problem with community.general. I will isolate this change in a new PR.

mattclay:
@goneri I don't expect there to be any issues upgrading to Python 3.9, but it would be a good idea to look at the generated Lambda zip files to make sure there are no unexpected new files. I also recommend reviewing the logs once this is deployed to make sure no unexpected issues show up.

tox.ini
mattclay: I'm curious, how does this change have any effect on community.general?

goneri: This was the error:

```
TASK [Fail if policy is too large] ***********************************************************************************************************************************************************************************************************
[WARNING]: an unexpected error occurred during Jinja2 environment setup: unable to locate collection community.general
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while templating '{{ policies | json_query(_query) | list }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: template error while templating string: unable to locate collection community.general. String: {{ policies | json_query(_query) | list }}"}
```

See: https://github.com/mattclay/aws-terminator/pull/250

goneri:
@mattclay I did the test in our team account (ca-central-1) before pushing this series of PR, it works fine.

mattclay (approved):

PR #250, opened 2022-11-29
tox: run policy by default

goneri:
- install the collections in `.tox/ansible`
- we need `community.general` because of `json_query`.


goneri:
We need https://github.com/mattclay/aws-terminator/pull/249 first.

gravesm (commented):
Could you add `.tox` to `.gitignore`?

goneri:
> Could you add `.tox` to `.gitignore`?

-> https://github.com/mattclay/aws-terminator/pull/252

gravesm (approved):

PR #254, opened 2022-12-16
Remove Security Group Rules when deletion failed with DependencyViolation exception

abikouo:
Update terminator Class Ec2SecurityGroup

### DIFF
--- a/aws/terminator/networking.py
+++ b/aws/terminator/networking.py
@@ -1,5 +1,5 @@
 import datetime
-
+import botocore
 from . import DbTerminator, Terminator, get_tag_dict_from_tag_list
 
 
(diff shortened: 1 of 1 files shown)


aws/terminator/networking.py
```diff
@@ -414,8 +414,23 @@ def name(self):
...
+        )
+
+        # Revoke Ingress rules
+        self.client.revoke_security_group_ingress(
+            GroupId=self.id, IpPermissions=self.instance.get("IpPermissions")
+        )
+
     def terminate(self):
-        self.client.delete_security_group(GroupId=self.id)
+        try:
```
goneri, line 429: What happen if we just call the two methods without the if condition? This way we avoid the risk of breakage if the format of the exception change in the future. I imagine we may need to wrap `revoke_sg_rules()` in `try/except` block thus.
```
self.revoke_sg_rules()
self.client.delete_security_group(GroupId=self.id)
```

goneri (approved):

aws/terminator/networking.py
abikouo: This was my first idea, but we want to cut down on unnecessary requests

jillr:
The code looks fine but I'm having trouble reproducing the original failure with the eks_cluster tests.  (I'm commenting out the SG and rule cleanup tasks at the end of full_test.yml then letting the terminator handle the SGs).  I'll test more tomorrow, the tests are a bit slow and it's getting late.  :)

jillr (approved):
This time I was able to reproduce the failure and confirm that this branch solves the issue.  LGTM, deploying.

PR #256, opened 2023-01-26
Allow cloudformation:DescribeType

alinabuzachis:
Signed-off-by: Alina Buzachis <abuzachis@redhat.com>

### DIFF
--- a/aws/policy/application-services.yaml
+++ b/aws/policy/application-services.yaml
@@ -10,6 +10,7 @@ Statement:
       - cloudformation:CreateResource
       - cloudformation:DeleteResource
       - cloudformation:DescribeStacks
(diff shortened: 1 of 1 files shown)


aws/policy/application-services.yaml
```diff
@@ -16,6 +16,7 @@ Statement:
       - cloudformation:ListResourceRequests
       - cloudformation:ListResources
       - cloudformation:UpdateResource
+      - cloudformation:DescribeType
```
gravesm, line 19: Could you make sure this change is ordered alphabetically? It starts to get difficult to manage if things get all mixed in.

gravesm (approved):

PR #257, opened 2023-01-26
Add terminator cleanup for SSM sessions

tremble:


gravesm (commented):
It looks like this will also need to add the `ssm:DescribeSessions` permission.

gravesm (approved):

PR #264, opened 2023-02-16
ECS Policies

tremble:
Builds on top of work done by alinabuzachis in #210 

Tested using https://github.com/ansible-collections/community.aws/pull/1716


gravesm:
Thanks, I'll start taking a look at this and see if we can finally get this one merged.

aws/policy/security-services.yaml
```diff
@@ -31,6 +31,7 @@ Statement:
           - 'arn:aws:iam::aws:policy/service-role/AmazonDMSVPCManagementRole'
           - 'arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole'
           - 'arn:aws:iam::aws:policy/service-role/AWSServiceRoleForVPCTransitGateway'
+          - 'arn:aws:iam::aws:policy/service-role/AmazonEC2ContainerServiceRole'
```
gravesm, line 34: This one is duplicated, and probably should be `arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy`

aws/terminator/paas.py
```diff
@@ -117,3 +117,131 @@ def terminate(self):
             else:
                 # delete streaming distribution
                 self.client.delete_streaming_distribution(Id=self.Id, IfMatch=ETag)
+
+
+class Ecs(DbTerminator):
+    @property
+    def age_limit(self):
+        return datetime.timedelta(minutes=20)
```
jillr, line 125: The current import causes this to fail.  I'd suggest this change earlier in the file:
```
+++ b/aws/terminator/paas.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+import datetime
 
 from . import DbTerminator, Terminator
 
@@ -39,7 +39,7 @@ class LambdaLayers(Terminator):
 
     @property
     def created_time(self):
-        return datetime.strptime(self.instance['LatestMatchingVersion']['CreatedDate'], "%Y-%m-%dT%H:%M:%S.%f%z")
+        return datetime.datetime.strptime(self.instance['LatestMatchingVersion']['CreatedDate'], "%Y-%m-%dT%H:%M:%S.%f%z")
 
     def terminate(self):
         for version in self.client.list_layer_versions(LayerName=self.name)['LayerVersions']:
```

aws/policy/security-services.yaml
```diff
@@ -204,3 +197,4 @@ Statement:
           - 'eks-nodegroup.amazonaws.com'
           - 'transitgateway.amazonaws.com'
           - 'network-firewall.amazonaws.com'
+          - 'ecs.amazonaws.com'
```
jillr, line 200: In addition to the change @gravesm suggested, we also need:
```suggestion
          - 'ecs.amazonaws.com'
          - 'ecs-test.amazonaws.com'
```
This will push the file over the character limit even with your wildcard changes, #266 will help with that.

jillr (changes requested):

gravesm (approved):

jillr (approved):

PR #266, opened 2023-02-21
Remove ansible_lambda_role policy

jillr:
We're no longer testing c.aws on Ansible 2.9 and policy/security-services.yaml is very close to the 6144 character limit.  We will also need to delete the role on the CI account.


tremble (approved):

gravesm:
Closing this as the change was incorporated into #264.

PR #267, opened 2023-02-24
Add permissions for Lightsail Instance Snapshots and change public ports

Nfsaavedra:
Allows integrations tests for creation and deletion of Lightsail Instance Snapshots.
Allows integrations tests for changing the public ports of Lightsail Instances.
Need it for ansible-collections/community.aws#1724.


markuman (approved):

aws/policy/paas.yaml
```diff
@@ -94,6 +94,10 @@ Statement:
       - lightsail:RebootInstance
       - lightsail:StopInstance
       - lightsail:ReleaseStaticIp
+      - lightsail:PutInstancePublicPorts
+      - lightsail:GetInstanceSnapshot
+      - lightsail:CreateInstanceSnapshot
```
gravesm, line 99: Could you move the `lightsail:CreateInstanceSnapshot` up to the `AllowResourceRestrictedActionsWhichIncurFees` Sid? And can you also add `lightsail:GetInstanceSnapshots` here? Otherwise, looks good.

Nfsaavedra: I did not notice the other Sid. Its done!

gravesm (approved):

PR #290, opened 2023-08-07
Refactor policies to decrease their length

alinabuzachis:
Refactor policies to decrease their length

### DIFF
--- a/aws/policy/application-services.yaml
+++ b/aws/policy/application-services.yaml
@@ -14,11 +14,22 @@ Statement:
       - cloudformation:List*
       - cloudformation:UpdateResource
       ###
+      - cloudwatch:Describe*
       - codebuild:BatchGetProjects
+      - codebuild:CreateProject
+      - codebuild:DeleteProject
       - codebuild:List*
+      - codebuild:UpdateProject
+      - codecommit:CreateRepository
+      - codecommit:DeleteRepository
+      - codecommit:Get*
       - codecommit:List*
+      - codecommit:UpdateRepositoryDescription
+      - codepipeline:CreatePipeline
+      - codepipeline:DeletePipeline
       - codepipeline:Get*
       - codepipeline:List*
+      - codepipeline:UpdatePipeline
       - ec2messages:AcknowledgeMessage
       - ec2messages:DeleteMessage
       - ec2messages:FailMessage
@@ -34,6 +45,10 @@ Statement:
       - glue:Get*
       - kinesis:Describe*
       - kinesis:List*
+      - mq:CreateBroker
+      - mq:CreateTags
+      - mq:DeleteBroker
+      - mq:Describe*
       - mq:List*
       - ses:CreateReceiptRuleSet
       - ses:DeleteIdentity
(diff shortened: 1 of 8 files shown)


aws/policy/networking.yaml
```diff
@@ -88,9 +79,7 @@ Statement:
       - ec2:DeleteTransitGateway
       - ec2:DeleteTransitGatewayPeeringAttachment
       - ec2:DeleteTransitGatewayVpcAttachment
-      - ec2:DescribeSubnets
-      - ec2:DescribeVpcs
-      - ec2:DescribeSecurityGroups
+      - ec2:Describe*
```
gravesm, line 82: This conflicts with the `ec2:Describe*` policy in `compute.yaml`

aws/policy/networking.yaml
```diff
@@ -132,7 +121,7 @@ Statement:
       - network-firewall:DeleteFirewallPolicy
       - network-firewall:DeleteResourcePolicy
       - network-firewall:DeleteRuleGroup
-      - network-firewall:ListTagsForResource
+      - network-firewall:List*
```
gravesm, line 124: Conflicts with the `network-firewall:List*` above.

gravesm (commented):
I added a few comments here about conflicting policies, but didn't do that for every file. I think just about every file in this PR has conflicts that will need to be resolved in some way. It may be that for many of them, we can just relax the resource restriction and move those up to the relevant unrestricted policy. Some we may need to keep in resource restricted. Keep in mind that since we don't really use `Deny`, a liberal `Allow` policy like `ecr:Get*` on unrestricted resources will functionally override a more restrictive `Allow` policy like `ecr:GetAuthorizationToken` on a resource restricted policy.

aws/policy/application-services.yaml
```diff
@@ -102,105 +121,44 @@ Statement:
...
-      - cloudformation:GetStackPolicy
-      - cloudformation:GetTemplate
-      - cloudformation:List*
       - cloudformation:SetStackPolicy
       - cloudformation:UpdateStack
       - cloudformation:UpdateTerminationProtection
       - cloudwatch:DeleteAlarms
-      - cloudwatch:DescribeAlarms
       - cloudwatch:PutMetricAlarm
-      - codebuild:CreateProject
```
alinabuzachis, line 117: They are nor resource restricted right now. I moved these into resource unrestricted, but if we want to restrict those, I can add the restrictions (same for codecommit, sns, states).

aws/policy/data-services.yaml
```diff
@@ -25,15 +28,13 @@ Statement:
...
-      - dms:ListTagsForResource
       - dms:ModifyEndpoint
       - dms:ModifyReplicationSubnetGroup
       - dms:RemoveTagsFromResource
       - dynamodb:CreateTable
       - dynamodb:DeleteItem
       - dynamodb:DeleteTable
-      - dynamodb:DescribeContinuousBackups
-      - dynamodb:GetItem
+      - dynamodb:Describe*
```
gravesm, line 37: This is covered above and can be removed.

aws/policy/application-services.yaml
gravesm: Can we keep all the actions that create, modify or delete resources in this sid? They aren't really resource restricted, but they are region restricted. I'm not sure why it's done this way, but it may be that some of these don't support the `RequestedRegion` condition. If they do, we could move these to a different region restricted sid and remove the individual resource restrictions.

jillr (commented):
Aside from the already noted comments I think this looks good - thanks @alinabuzachis!

aws/policy/application-services.yaml
alinabuzachis: Done! I hope I haven't forgotten anything!

gravesm (commented):
One small thing, but otherwise looks good.

aws/policy/security-services.yaml
```diff
@@ -116,11 +112,10 @@ Statement:
...
       - acm:AddTagsToCertificate
       - acm:DeleteCertificate
-      - acm:DescribeCertificate
-      - acm:GetCertificate
+      - acm:Describe*
+      - acm:Get*
       - acm:ImportCertificate
-      - acm:ListCertificates
-      - acm:ListTagsForCertificate
+      - acm:List*
```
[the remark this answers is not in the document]
alinabuzachis: Done! Thanks.

gravesm (approved):

jillr (approved):

jillr:
I'll deploy this change asap, thanks @alinabuzachis!

PR #293, opened 2023-12-15
EC2: add policy for modifying instance metadata options

mandar242:
Added missing policies to modify ec2 instance metadata options.

PR needing permissions: https://github.com/ansible-collections/amazon.aws/pull/1918

API reference [ModifyInstanceMetadataOptions](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyInstanceMetadataOptions.html)


gravesm (approved):
There's an error in the tests, but this PR does fix the permissions issue.

PR #294, opened 2023-12-19
add policy s3:DeleteObjectVersion

abikouo:
Trying to delete a non-empty bucket created with `versioning=true` required the following permission `s3:DeleteObjectVersion` as the module `amazon.aws.s3_bucket` will try to delete each object versioning before deleting the bucket.

Required for https://github.com/redhat-cop/cloud.terraform_ops/pull/1


jillr (approved):

jillr:
I'm unable to fully test this as the terraform tasks in the linked PR fail but the policy change will be needed regardless so deploying to prod.  If additional policies are needed please open additional PRs.

PR #295, opened 2024-01-10
Increase max session duration to two hours

gravesm:


alinabuzachis (approved):

gravesm:
Yeah, this is mostly for the benefit of downstream testing which handles its own session creation.

PR #303, opened 2024-05-23
add secretsmanager:GetRandomPassword

markuman:
Required by https://github.com/ansible-collections/community.aws/pull/1997

### DIFF
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -106,6 +106,7 @@ Statement:
       - secretsmanager:PutResourcePolicy
       - secretsmanager:DeleteResourcePolicy
       - secretsmanager:RemoveRegionsFromReplication
+      - secretsmanager:GetRandomPassword
     Resource:
       - 'arn:aws:iam::{{ aws_account_id }}:server-certificate/ansible-test-*'
       - 'arn:aws:secretsmanager:{{ aws_region }}:{{ aws_account_id }}:secret:ansible-test*'
--- a/aws/policy/security-services.yaml
+++ b/aws/policy/security-services.yaml
@@ -88,6 +88,7 @@ Statement:
       - logs:List*
       - secretsmanager:Describe*
       - secretsmanager:List*
+      - secretsmanager:GetRandomPassword
     Resource: "*"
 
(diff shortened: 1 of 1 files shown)


markuman:
oh damn

```
TASK [Fail if policy is too large] *********************************************
fatal: [localhost]: FAILED! => {
    "failed_when_result": true,
    "oversized_policies": [
        "/home/vsts/work/1/s/hacking/aws_config/../../aws/policy/security-services.yaml"
    ]
}
```

aws/policy/security-services.yaml
```diff
@@ -106,6 +106,7 @@ Statement:
       - secretsmanager:PutResourcePolicy
       - secretsmanager:DeleteResourcePolicy
       - secretsmanager:RemoveRegionsFromReplication
+      - secretsmanager:GetRandomPassword
```
gravesm, line 109: I think the problem is that this policy statement is resource restricted and the `GetRandomPassword` permission does not operate on a secret. I just tested by adding this permission to the `AllowGlobalUnrestrictedResourceActionsWhichIncurNoFees` policy statement and the associated PR passes. Unfortunately, this policy is right at the maximum allowed characters and there's not a lot here that can be collapsed. About the only thing I can see would be to replace `kms:DisableKey` and `kms:DisableKeyRotation` with `kms:Disable*`. There are only two `kms:Disable*` permissions so this won't add anything new.

markuman: Sounds good. I've reduced the two rules to `kms:Disable*`.

aws/policy/security-services.yaml
```diff
@@ -106,6 +105,7 @@ Statement:
...
```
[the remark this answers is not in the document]
markuman: done

aws/policy/security-services.yaml
```diff
@@ -89,6 +88,7 @@ Statement:
       - logs:List*
       - secretsmanager:Describe*
       - secretsmanager:List*
+      - secretsmanager:GetRandomPassword
```
gravesm, line 91: Small nit, but could you make sure this is ordered alphabetically?

markuman: sure.

gravesm (approved):

PR #304, opened 2024-07-11
Include t3a.micro in compute.yaml

GomathiselviS:
This PR adds t3a.micro instance type to aws/policy/compute.yaml

Refer: https://issues.redhat.com/browse/ACA-1691

### DIFF
--- a/aws/policy/compute.yaml
+++ b/aws/policy/compute.yaml
@@ -26,6 +26,7 @@ Statement:
           - t2.micro
           - t3.nano
           - t3.micro
+          - t3a.micro
           - m1.large  # lowest cost instance type with EBS optimization supported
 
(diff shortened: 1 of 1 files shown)


tremble (approved):
This is a sensible change:
t3a.micro  (0.9¢/h)
t3.micro (~1¢/h)

alinabuzachis (approved):

gravesm (commented):
The compute policy is too big. At this point, there's nothing obvious to trim down. I think I would recommend replacing all the delete permissions for `autoscaling` in the `AllowGlobalResourceRestrictedActionsWhichIncurNoFees` with `autoscaling:Delete*`. The only new permission this adds is the ability to delete warm pools, and I think that's fine.

PR #307, opened 2024-09-04
Add rds:CreateTenantDatabase permission

GomathiselviS:
This PR adds rds:CreateTenantDatabase permission.

### DIFF
--- a/aws/policy/data-services.yaml
+++ b/aws/policy/data-services.yaml
@@ -78,6 +78,7 @@ Statement:
       - rds:CreateTenantDatabase
       - rds:ModifyDBInstance
       - rds:DeleteDBInstance
+      - rds:DeleteTenantDatabase
       - rds:StopDBCluster
       - rds:StopDBInstance
       - rds:StartDBCluster
@@ -125,6 +126,7 @@ Statement:
       - 'arn:aws:dms:{{ aws_region }}:{{ aws_account_id }}:endpoint:*'
       - 'arn:aws:rds:{{ aws_region }}:{{ aws_account_id }}:snapshot:*'
       - 'arn:aws:rds:{{ aws_region }}:{{ aws_account_id }}:cluster-snapshot:*'
+      - ''arn:aws:rds:{{ aws_region }}:{{ aws_account_id }}:tenant-database:*'
   - Sid: AllowGlobalRestrictedResourceActionsWhichIncurFees
     Effect: Allow
     Action:


alinabuzachis (approved):

aws/policy/data-services.yaml
```diff
@@ -75,6 +75,7 @@ Statement:
       - rds:RestoreDBInstanceFromS3
       - rds:CreateDBInstanceReadReplica
       - rds:CreateDBInstance
+      - rds:CreateTenantDatabase
```
abikouo, line 78:
```suggestion
      - rds:*TenantDatabase
```
to support `ModifyTenantDatabase` and `DeleteTenantDatabase`

abikouo (changes requested):

aws/policy/data-services.yaml
gravesm: We should avoid wildcarding in this way. If there's a need to reduce the size of this policy there are better ways to do it.

alinabuzachis: I was about to say that. We add wildcarding for List or Get permissions which are harmless. I'm testing right now with the terminator and I would only stick to the permissions needed for that PR.

alinabuzachis (commented):
@GomathiselviS Creating an oracle-ee-cdb DB instance is very slow. I think I waited more than 30 minutes for this `TASK [rds_instance_modify : Create an oracle-ee-cdb DB instance with multi tenant option]. ***` (I did not measure the time though). I would suggest adding unit tests for this option or try adding integration tests for this feature in a dedicated test role because it would heavily impact on the execution time of the rds_instance_modify test role (this may bring  Zuut timeouts in the CI). Was the execution of this task quicker for you.

In any case you would also need to add `- 'arn:aws:rds:{{ aws_region }}:{{ aws_account_id }}:tenant-database:*'` to the Resource section. And you also need to add `DeleteTenantDatabase` permission.

gravesm:
I just want to confirm, I notice there are boto functions for tenant databases (for example, https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_tenant_databases.html). Does this need new terminator classes, or will the existing RDS terminators handle cleaning this up?

alinabuzachis:
> I just want to confirm, I notice there are boto functions for tenant databases (for example, https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_tenant_databases.html). Does this need new terminator classes, or will the existing RDS terminators handle cleaning this up?

Still trying to check that because the first time it failed because of the DeleteTenantOption missing. So, I'm waiting for the second run to complete and check.

aws/policy/data-services.yaml
alinabuzachis: @gravesm  It seems to me what we do not need a dedicated class. I tried to force the cleanup.

```
(py311) alinabuzachis@Alinas-MacBook-Pro aws % python3 cleanup.py  --stage dev --target RdsDbInstance --force

cleanup     : INFO     terminated RdsDbInstance: name=ansible-test-oracle-339547aee27a, id=arn:aws:rds:us-east-1:xxx:db:ansible-test-oracle-339547aee27a age=0:03:31, stale=False
cleanup     : INFO     terminated RdsDbInstance: name=ansible-test-oracle-8581d1fe8fa7, id=arn:aws:rds:us-east-1:xxx:db:ansible-test-oracle-8581d1fe8fa7 age=0:03:30, stale=False
```
And if I re-run the same command I get for both
`.InvalidDBInstanceStateFault: An error occurred (InvalidDBInstanceState) when calling the ModifyDBInstance operation: The specified database instance is currently in deleting state, which is a terminal state."}.InvalidDBInstanceStateFault: An error occurred (InvalidDBInstanceState) when calling the ModifyDBInstance operation: The specified database instance is currently in deleting state, which is a terminal state."}`

Then after some time:
```
(py311) alinabuzachis@Alinas-MacBook-Pro aws % python3 cleanup.py  --stage dev --target RdsDbInstance --force -vvvv

cleanup     : DEBUG    Config path: /Users/alinabuzachis/dev/aws-terminator/aws/config.yml
cleanup     : DEBUG    located RdsDbInstance: count=0
```

I also tried:
```
aws rds describe-tenant-databases --region us-east-1     
{
    "TenantDatabases": []
}
```

alinabuzachis: It seems @GomathiselviS has added some units tests for the functionality https://github.com/ansible-collections/amazon.aws/pull/2275 and removed the integration tests. @gravesm Do we still want to add merge this change or can we close this PR? I'll be inclined to close it.

GomathiselviS:
Closing this ticket, as we do not have integration tests (replaced with unit tests) that would need this permission.

PR #311, opened 2024-10-07
Additional ASG permissions for autoscaling_instance

tremble:
Needed for ~https://github.com/ansible-collections/amazon.aws/pull/2296~ https://github.com/ansible-collections/amazon.aws/issues/2362

### DIFF
--- a/aws/policy/compute.yaml
+++ b/aws/policy/compute.yaml
@@ -131,7 +131,7 @@ Statement:
     Action:
       - autoscaling:EnableMetricsCollection
       - ec2:CreateVolume
-      - elasticloadbalancing:CreateLoadBalancer
+      - elasticloadbalancing:CreateLoadBalancer*
       - elasticloadbalancing:CreateRule
     Resource:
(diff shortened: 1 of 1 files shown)


aws/policy/compute.yaml
gravesm, line 1: Looks like this also needs at least `autoscaling:ResumeProcesses` and `autoscaling:SuspendProcesses`. Still failing with these, but not for permission reasons.

alinabuzachis, line 1: It seems there's an issue with the length. I'll probably suggest we do something:

```
- elasticloadbalancing:CreateLoadBalancerListeners
- elasticloadbalancing:CreateLoadBalancerPolicy
```
to

```
- elasticloadbalancing:CreateLoadBalancer*
```

We could also do something similar for:

```
- elasticloadbalancing:SetLoadBalancerPoliciesForBackendServer
 - elasticloadbalancing:SetLoadBalancerPoliciesOfListener
```

to 

```
 - elasticloadbalancing:SetLoadBalancerPolicies*
```

GomathiselviS (approved):

PR #330, opened 2026-02-12
Add Service Discovery permissions and terminator for ecs_service

jonpspri:
## Summary
- Expand `servicediscovery` IAM permissions to support all namespace types (HTTP, Public DNS, Private DNS) along with `GetNamespace`, `ListServices`, and `TagResource`
- Add ELB service-linked role creation permission to the paas policy
- `ServiceDiscoveryNamespace` terminator to clean up all namespace types, including deleting child services before namespace removal

## Context
These changes support the `ecs_service` and `cloudmap` integration tests in community.aws:
- https://github.com/ansible-collections/community.aws/pull/2408
- https://github.com/ansible-collections/community.aws/pull/2411

## Test plan
- [x] Verify terminator discovers and cleans up all namespace types (HTTP, DNS_PUBLIC, DNS_PRIVATE) in the test account
- [x] Verify child services are deleted before namespace removal
- [X] Verify IAM policies deploy without errors


aws/terminator/paas.py
```diff
@@ -248,15 +248,6 @@ def _paginate_task_results(container_instance=None):
 
             return [] if not names else names
 
-        def _paginate_task_definition_results():
```
alinabuzachis, line 251: I don’t believe this should be removed, as there is a dependency hierarchy when ECS resources are cleaned up.

aws/policy/paas-ecs.yaml
```diff
@@ -0,0 +1,38 @@
+Version: '2012-10-17'
```
alinabuzachis: I don’t believe this split would resolve the issue. If I’m not mistaken, the main constraint is the Lambda package size limit/policy size limit, which is why each file has its current size. Splitting the files wouldn’t address that limit. When we approach the limit, we typically refactor instead.

@gravesm do you have any suggestion?

aws/policy/networking.yaml
```diff
@@ -22,6 +22,10 @@ Statement:
       - route53:UpdateHealthCheck
       - network-firewall:List*
       - network-firewall:Describe*
+      - servicediscovery:CreateHttpNamespace
+      - servicediscovery:CreatePublicDnsNamespace
+      - servicediscovery:CreatePrivateDnsNamespace
+      - servicediscovery:ListNamespaces
```
gravesm: I believe these should be region restricted.

aws/policy/networking.yaml
```diff
@@ -163,3 +167,15 @@ Statement:
...
+
+  - Sid: AllowServiceDiscoveryActionsWhichIncureNoFees
+    Effect: Allow
+    Action:
+      - servicediscovery:DeleteNamespace
+      - servicediscovery:GetNamespace
+      - servicediscovery:GetOperation
+      - servicediscovery:ListServices
+      - servicediscovery:DeleteService
+      - servicediscovery:TagResource
```
gravesm: Can you move these up to the existing region restricted sid?

aws/terminator/paas.py
jonpspri: The nested deletion of TaskDefinitions was removed from the Ecs terminator due to a blast radius issue.

The problem was that when cleaning up any stale ECS cluster, the old code would deregister all task definitions in the entire account - not just the ones associated with that specific cluster. Looking at the old code:

**Old code in Ecs.terminate()**
```
  task_definitions = _paginate_task_definition_results()  # Gets ALL task definitions
  for each in task_definitions:
      self.client.deregister_task_definition(taskDefinition=each)
```

This `_paginate_task_definition_results()` function listed all task definitions account-wide without any filtering by cluster. So whenever any cluster was cleaned up, it would deregister task definitions that might still be in use by other clusters or services.

The fix was to:
  1. Extract task definition cleanup into its own `EcsTaskDefinition(DbTerminator)` class
  2. Track task definitions independently via DynamoDB with their own 20-minute age limit
  3. Clean them up based on their own lifecycle, not as a side effect of cluster cleanup
  4. Also properly delete (not just deregister) stale task definitions

This way, task definitions are only cleaned up when they themselves are stale, rather than being caught in the blast radius of any cluster cleanup.

aws/policy/paas-ecs.yaml
jonpspri: The concern is actually the policy size limit of 6144 characters.  I've tested with the split and it works.

alinabuzachis: @gravesm was suggesting to do something like:
``` 
--- a/aws/policy/paas.yaml
+++ b/aws/policy/paas.yaml
@@ -40,10 +40,7 @@ Statement:
       - cloudfront:CreateInvalidation
       - cloudfront:CreateOriginRequestPolicy
       - cloudfront:DeleteCloudFrontOriginAccessIdentity
-      - cloudfront:DeleteCachePolicy
-      - cloudfront:DeleteDistribution
-      - cloudfront:DeleteOriginRequestPolicy
-      - cloudfront:DeleteStreamingDistribution
+      - cloudfront:Delete*
       - cloudfront:TagResource
       - cloudfront:UntagResource
       - cloudfront:UpdateCachePolicy
```

By doing this and moving the other policies into the pass.yml file (from the file paas-ecs.yml), everything should be fine.

aws/policy/networking.yaml
```diff
@@ -163,3 +167,15 @@ Statement:
     Resource:
       - 'arn:aws:apigateway:{{ aws_region }}::/restapis*'
       - 'arn:aws:apigateway:{{ aws_region }}::/tags/*'
+
+  - Sid: AllowServiceDiscoveryActionsWhichIncurNoFees
+    Effect: Allow
+    Action:
+      - servicediscovery:DeleteNamespace
+      - servicediscovery:GetNamespace
+      - servicediscovery:GetOperation
```
alinabuzachis:
```suggestion
      - servicediscovery:Get*
```

aws/policy/paas.yaml
```diff
@@ -36,20 +36,11 @@ Statement:
...
       - bedrock:UpdateAgent*
-      - cloudfront:CreateCachePolicy
-      - cloudfront:CreateInvalidation
-      - cloudfront:CreateOriginRequestPolicy
-      - cloudfront:DeleteCloudFrontOriginAccessIdentity
-      - cloudfront:DeleteCachePolicy
-      - cloudfront:DeleteDistribution
-      - cloudfront:DeleteOriginRequestPolicy
-      - cloudfront:DeleteStreamingDistribution
+      - cloudfront:Create*
```
alinabuzachis: I don't think we should wildcard Create and Update operations.

aws/policy/paas.yaml
```diff
@@ -39,17 +39,11 @@ Statement:
...
-      - cloudfront:DeleteDistribution
-      - cloudfront:DeleteOriginRequestPolicy
-      - cloudfront:DeleteStreamingDistribution
+      - cloudfront:Delete*
       - cloudfront:TagResource
       - cloudfront:UntagResource
       - cloudfront:UpdateCachePolicy
       - cloudfront:UpdateCloudFrontOriginAccessIdentity
-      - cloudfront:UpdateDistribution
-      - cloudfront:UpdateOriginRequestPolicy
```
alinabuzachis, line 52: Why are these two removed?

aws/policy/networking.yaml
```diff
@@ -136,11 +136,22 @@ Statement:
...
+      - servicediscovery:CreateHttpNamespace
+      - servicediscovery:CreatePublicDnsNamespace
+      - servicediscovery:CreatePrivateDnsNamespace
+      - servicediscovery:ListNamespaces
+      - servicediscovery:DeleteNamespace
+      - servicediscovery:GetNamespace
+      - servicediscovery:GetOperation
+      - servicediscovery:ListServices
+      - servicediscovery:DeleteService
+      - servicediscovery:TagResource
```
gravesm, line 148: Can you sort these alphabetically?

PR #332, opened 2026-02-14
Add Cognito user pool policy and terminator

jonpspri:
## Summary

- Add IAM policy for `cognito-idp` operations (user pools, clients, domains, and managed login branding)
- Add `CognitoUserPool` terminator class to clean up stale user pools left by integration tests
- The terminator handles deleting user pool domains before the pool itself (required by AWS), while clients and branding are automatically cleaned up with the pool

## Related

Supports ansible-collections/community.aws#2412, which adds new Cognito modules (`cognito_user_pool`, `cognito_user_pool_client`, `cognito_user_pool_domain`, `cognito_managed_login_branding`) with integration tests that create user pool resources.

## Test plan

- [ ] Verify terminator lambda discovers and cleans up Cognito user pools
- [ ] Verify user pool domains are deleted before the pool
- [ ] Verify policy grants sufficient permissions for both integration tests and teardown

### DIFF
--- a/aws/policy/cognito.yaml
+++ b/aws/policy/cognito.yaml
@@ -1,0 +1,35 @@
+Version: '2012-10-17'
+Statement:
+
+  - Sid: CognitoServicesWhichDoNotIncurFees
+    Effect: Allow
+    Action:
+      - cognito-idp:CreateManagedLoginBranding
+      - cognito-idp:CreateUserClient
+      - cognito-idp:CreateUserPool
+      - cognito-idp:CreateUserPoolClient
+      - cognito-idp:CreateUserPoolDomain
+      - cognito-idp:DeleteManagedLoginBranding
+      - cognito-idp:DeleteUserPool
+      - cognito-idp:DeleteUserPoolClient
+      - cognito-idp:DeleteUserPoolDomain
+      - cognito-idp:DescribeManagedLoginBranding
+      - cognito-idp:DescribeManagedLoginBrandingByClient
+      - cognito-idp:DescribeUserPool
+      - cognito-idp:DescribeUserPoolClient
+      - cognito-idp:ListUserPoolClients
+      - cognito-idp:TagResource
+      - cognito-idp:UntagResource
+      - cognito-idp:UpdateManagedLoginBranding
+      - cognito-idp:UpdateUserPool
+      - cognito-idp:UpdateUserPoolClient
+    Resource:
+      - 'arn:aws:cognito-idp:{{ aws_region }}:{{ aws_account_id }}:userpool/*'
+
+  - Sid: CognitoServicesWhichDoNotIncurFeesGlobal
+    Effect: Allow
+    Action:
+      - cognito-idp:ListUserPools
+      - cognito-idp:DescribeUserPoolDomain
+    Resource:
+      - '*'
--- a/aws/policy/application-security.yaml
+++ b/aws/policy/application-security.yaml
@@ -85,3 +85,36 @@ Statement:
       StringEquals:
         aws:RequestedRegion:
           - '{{ aws_region }}'
+
+  - Sid: CognitoServicesWhichDoNotIncurFees
+    Effect: Allow
+    Action:
+      - cognito-idp:CreateManagedLoginBranding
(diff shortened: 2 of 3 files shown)


aws/policy/cognito.yaml
```diff
@@ -0,0 +1,36 @@
+---
+Version: '2012-10-17'
+Statement:
+
+  - Sid: CognitoServicesWhichDoNotIncurFees
+    Effect: Allow
+    Action:
+      - cognito-idp:CreateManagedLoginBranding
+      - cognito-idp:CreateUserClient
```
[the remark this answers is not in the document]
jonpspri: Testing failed on this permission, so I'm not believing the permission does not exist.

aws/policy/cognito.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - cognito-idp:CreateUserPool
+      - cognito-idp:CreateUserPoolClient
+      - cognito-idp:CreateUserPoolDomain
```
alinabuzachis, line 11:
```suggestion
      - cognito-idp:CreateUserPool*
```

aws/policy/cognito.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - cognito-idp:DeleteManagedLoginBranding
+      - cognito-idp:DeleteUserPool
+      - cognito-idp:DeleteUserPoolClient
+      - cognito-idp:DeleteUserPoolDomain
```
alinabuzachis, line 15: Maybe we could also wildcard - cognito-idp:Describe*, but you can also have:

- cognito-idp:DescribeManagedLoginBranding*
- cognito-idp:DescribeUserPool*

aws/terminator/cognito.py
```diff
@@ -0,0 +1,32 @@
...
+    @property
+    def name(self):
+        return self.instance['Name']
+
+    @property
+    def created_time(self):
+        return self.instance['CreationDate']
+
+    def terminate(self):
+        # User pool domains must be deleted before the user pool can be deleted
```
alinabuzachis: Shouldn't the ManagedLoginBranding be deleted too?

alinabuzachis (commented):
Could you please move these files into existing ones? Maybe application-services?

aws/policy/cognito.yaml
alinabuzachis, line 15:
```suggestion
      - cognito-idp:DeleteUserPool*
```

aws/terminator/cognito.py
jonpspri: ⏺ No, you likely don't need a separate ManagedLoginBranding section in the terminator. Here's my reasoning:

  1. ManagedLoginBranding is a sub-resource of a User Pool — it exists only within a user pool and has no independent lifecycle. When the user pool is
  deleted, its branding should be cleaned up automatically by AWS.
  2. No list API exists — Unlike user pools (which have list_user_pools), there's no ListManagedLoginBrandings API. You can only DescribeManagedLoginBranding
  by ID or DescribeManagedLoginBrandingByClient by client+pool ID, making it impractical to create a standalone Terminator subclass that discovers these
  resources.
  3. The domain precedent is different — The existing CognitoUserPool.terminate() at aws/terminator/cognito.py:24-32 explicitly deletes user pool domains
  before the pool because domains are globally unique DNS resources that must be released before pool deletion. ManagedLoginBranding doesn't have that
  constraint.

  If it turns out that delete_user_pool fails when ManagedLoginBranding is present (which would be surprising), the right fix would be to add cleanup in the
  existing CognitoUserPool.terminate() method — enumerate branding via ListUserPoolClients + DescribeManagedLoginBrandingByClient and delete each one before
  calling delete_user_pool — similar to the domain handling pattern already there.

aws/policy/cognito.yaml
jonpspri: Not worth it for now -- there's plenty of room in this file and I dislike wildcarding any write-capable permissions.

jonpspri: (In truth, I dislike wildcarding ANY permissions, but you do what you have to do.)

PR #342, opened 2026-08-19
Add SageMaker Image IAM permissions and terminator

JanLikar:
## Summary

Adds the IAM permissions and terminator needed for the `sagemaker_image` integration tests in [ansible-collections/amazon.ai](https://github.com/ansible-collections/amazon.ai) (see PR [#54](https://github.com/ansible-collections/amazon.ai/pull/54)).

### `aws/policy/paas.yaml`

- Added `sagemaker:CreateImage`, `sagemaker:DescribeImage`, `sagemaker:UpdateImage`, `sagemaker:DeleteImage` to `AllowResourceRestrictedActionsWhichIncurNoFees` with resource `arn:aws:sagemaker:...:image/*`
- Added `sagemaker:ListImages`, `sagemaker:AddTags`, `sagemaker:DeleteTags`, `sagemaker:ListTags` to `AllowUnrestrictedResourceActionsWhichIncurNoFees`

### `aws/policy/security-services.yaml`

- Added `arn:aws:iam::aws:policy/AmazonSageMakerFullAccess` to the allowed `iam:PolicyArn` condition for `iam:AttachRolePolicy` so the test IAM role can be set up correctly

### `aws/terminator/paas.py`

- Added `SageMakerImage` terminator class to clean up leaked SageMaker Image resources after test runs, following the same pattern as the existing `SageMakerCodeRepository` terminator

PR #343, opened 2026-08-20
Redistribute ECR policies and clean up policy files

alinabuzachis:
## Summary

  - **Redistribute ECR** from `paas.yaml` to `storage-services.yaml` — natural fit alongside S3, EFS, Backup, and MemoryDB, and frees up space in `paas.yaml` for future
  SageMaker/Bedrock growth
  - **Fix typo**: `memorydb:DeleleSnapshot` → `memorydb:DeleteSnapshot` (silent permission bug — delete-snapshot was never actually granted)
  - **Remove duplicates**: `glue:Get*` was in both `application-services.yaml` and `data-services.yaml`; duplicate `arn:aws:dms:...endpoint:*` resource ARN in
  `data-services.yaml`
  - **Sort alphabetically**: all action lists and resource ARNs across all 8 policy files, making it easier to spot duplicates and review diffs
  - **Wildcard read-only actions**: `lightsail:Get*`, `bedrock:Invoke*`, `cloudfront:CreateStreamingDistribution*`, `ecs:*CapacityProvider*`
  - **Remove redundancy**: `eks:DescribeNodegroup` was already covered by `eks:Describe*`

  ## Policy size impact (JSON chars, limit: 6,144)

  | Policy file | Before | After | Change | Headroom |
  |---|---|---|---|---|
  | paas.yaml | 6,021 | 5,117 | -904 (-15.0%) | 1,027 |
  | storage-services.yaml | 2,749 | 3,548 | +799 (+29.1%) | 2,596 |
  | application-services.yaml | 4,818 | 4,805 | -13 (-0.3%) | 1,339 |
  | data-services.yaml | 5,433 | 5,384 | -49 (-0.9%) | 760 |
  | application-security.yaml | 2,182 | 2,182 | 0 | 3,962 |
  | compute.yaml | 5,791 | 5,791 | 0 | 353 |
  | networking.yaml | 5,420 | 5,420 | 0 | 724 |
  | security-services.yaml | 6,082 | 6,082 | 0 | 62 |

  Sizes calculated using the same method as `deploy-test-policy.yml`: `lookup('template', ...) | from_yaml | to_json | length`.

  ## Motivation

  PR #342 proposes adding 2 new policy files (`paas-sagemaker.yaml`, `security-services-slr.yaml`), which would bring the total to 10 — the AWS default limit for managed policies per role. This redistribution avoids that by freeing up ~900 chars in `paas.yaml` through moving ECR to `storage-services.yaml`, keeping the policy count at 8 and leaving room for future growth.

  ## Test plan
  
  - [ ] Verify all policies stay under 6,144 chars: `make test_policy STAGE=dev`
  - [ ] Verify no functional permission changes beyond the fixes noted above
  - [ ] Confirm ECR operations still work (permissions moved, not removed)
  - [ ] Confirm MemoryDB snapshot deletion works (typo fix grants the permission for the first time)

### DIFF
--- a/aws/policy/paas.yaml
+++ b/aws/policy/paas.yaml
@@ -10,13 +10,9 @@ Statement:
       - eks:CreateNodegroup
       - eks:UpdateNodegroupConfig
       - lambda:InvokeFunction
+      - lightsail:CreateInstanceSnapshot
       - lightsail:CreateInstances
       - lightsail:StartInstance
-      - lightsail:CreateInstanceSnapshot
-      - ecr:CompleteLayerUpload
-      - ecr:InitiateLayerUpload
-      - ecr:PutImage
-      - ecr:UploadLayerPart
     Resource:
       - 'arn:aws:bedrock:{{ aws_region }}:{{ aws_account_id }}:agent-alias/*'
       - 'arn:aws:bedrock:{{ aws_region }}::foundation-model/*'
@@ -24,7 +20,6 @@ Statement:
       - 'arn:aws:eks:{{ aws_region }}:{{ aws_account_id }}:nodegroup/*/*/*'
       - 'arn:aws:lambda:{{ aws_region }}:{{ aws_account_id }}:function:*'
       - 'arn:aws:lightsail:{{ aws_region }}:{{ aws_account_id }}:*'
-      - 'arn:aws:ecr:{{ aws_region }}:{{ aws_account_id }}:repository/*'
 
   - Sid: AllowResourceRestrictedActionsWhichIncurNoFees
     Effect: Allow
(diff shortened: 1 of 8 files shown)


oraNod (approved):
LGTM. Thanks @alinabuzachis

aws/policy/storage-services.yaml
```diff
@@ -66,6 +66,11 @@ Statement:
       - backup:UntagResource
       - backup:UpdateBackupPlan
       - backup-storage:MountCapsule
+      - ecr:CreateRepository
+      - ecr:Describe*
+      - ecr:GetAuthorizationToken
```
tremble: I think we can move to ecr:Get*

aws/policy/paas.yaml
```diff
@@ -136,24 +112,18 @@ Statement:
...
+      - cloudfront:CreateStreamingDistribution*
     Resource:
       - "*"
 
   - Sid: AllowUnrestrictedResourceActionsWhichIncurNoFees
     Effect: Allow
     Action:
-      - bedrock:ListFoundationModels
       - bedrock:GetFoundationModel
+      - bedrock:ListFoundationModels
```
tremble: I think we can move this to 
bedrock:Get*
bedrock:List*

tremble (approved):
As a follow-up I think we need to improve the documentation around these policies.

- Where do specific policies live (some services could live in multiple files)
- When is it "ok" to use wildcards (read only; does it need to be resource limited?)

Pointing Claude at it, also raised an interesting point:

- should we add a linter which enforces a sort order?

alinabuzachis:
> As a follow-up I think we need to improve the documentation around these policies.
> 
> * Where do specific policies live (some services could live in multiple files)
> * When is it "ok" to use wildcards (read only; does it need to be resource limited?)
> 
> Pointing Claude at it, also raised an interesting point:
> 
> * should we add a linter which enforces a sort order?

Thanks for pointing this out! I’ll work on adding more thorough documentation in an upcoming PR. Regarding the linter suggestion, that makes complete sense. I'm planning to do more work on this repo.

PR #353, opened 2026-09-12
Fix Ec2Eip and Ec2Eni termination errors for in-use resources

alinabuzachis:
Ec2Eip: Catch InvalidIPAddress.InUse when releasing EIPs still associated in non-default VPCs instead of logging an error every cycle.

Ec2Eni: Skip ENIs in associated/attaching/in-use/detaching states via ignore, allowing the parent resource to be cleaned up first.


oraNod (approved):
Thanks @alinabuzachis

